PR #54, opened 2023-04-03
amy.c: atoff() does not consume 'e'/'E' in messages.

dpwe:
Replace the atof() used in parsing Amy messages with atoff(), which does the same string -> float conversion, but does not recognize "scientific" format (1.2e7 etc.).  This will allow us to use "e" and "E" as command separators in the Amy wire protocol.  

Note that up until now, it worked to use scientific format in Amy messages, so this is a potentially non-backward-compatible change if anyone is using this format.  There's no real argument for using scientific format in these messages, though.

(I tried to submit this before, not sure what happened.)

### DIFF
--- a/extmod/amy/src/amy.c
+++ b/extmod/amy/src/amy.c
@@ -880,6 +880,26 @@ void parse_algorithm(struct event * e, char *message) {
 
 }
 
+float atoff(const char *s) {
+    // Returns float value corresponding to parseable prefix of s.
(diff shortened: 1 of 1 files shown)


bwhitman:
I will put this in Tulip, thank you dan!
Before i backport it to alles/AMY i need to make sure it doesn't cause any problems with parsing speed . i can't imagine how but you never know

PR #61, opened 2023-05-04
Added bg_noise() to modtulip.c and tulip_api.md

drepetto:
Fast background noise generator with various options, can be called at high framerates for fuzzy background.

### DIFF
--- a/docs/tulip_api.md
+++ b/docs/tulip_api.md
@@ -1,4 +1,4 @@
-# The Tulip Creative Computer API 
+# The Tulip Creative Computer API
 
 Here you can see the API [Tulip](../README.md) currently ships with.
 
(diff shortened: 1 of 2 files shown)


bwhitman:
This is great. I need to test this on real hardware (I'm away from it at the moment.)

I also would love this to be like a rect, where you can optionally set x/y/w/h 

Lastly, I wonder if an optimization for noise uses the "GPU" code in `display.c` `fill_bounce_buffer` instead of doing it in a double iteration. But I don't have a good idea there yet.

drepetto:
I think the thing that makes this complicated to optimize is the need to
randomly change the color of arbitrary sized blocks. One idea might be to
make one block_size x block_size block for each color and then randomly
memcopy those all over the screen. The problem might then be making sure
you touch every pixel. Maybe you do a total fill with one of the colors
first?

drepetto:
Or maybe you just do a single iteration over the whole image, incrementing
by block_size, and randomly picking one of the block_size x block_size
chunks to memcopy at each step. Something like that should work I think.
Not sure it'll actually be faster?

PR #65, opened 2023-05-08
Function in tulip.py to extract .tar files.

drepetto:
A start at tar/untar functionality...


bwhitman:
Nice! Thanks @drepetto

PR #73, opened 2023-05-13
Link local amy/readme.md to tulip_api.md

dpwe:
The local version of AMY includes features not present in the upstream one.  Make sure users can find the full AMY API documentation from the Tulip repo.


bwhitman:
Sad you didn't add the Venn diagram 
IMG_0800

PR #89, opened 2023-05-26
Fix overflow parsing long decimals in alles/amy.

dpwe:
Fixes issue where random pans were being scrambled.


dpwe:
Note that a maliciously-constructed raw message to AMY with a long decimal part would still overflow atoi.

bwhitman:
Thanks Dan!!

PR #90, opened 2023-05-26
amy: Flip sense of pan so 0=left, 1=right.

dpwe:
This is to conform to Douglas's expectations, and the convention used in Sound's pan2.


bwhitman:
image

PR #95, opened 2023-06-19
ports/esp32s3/usb_keyboard: Add support for USB-MIDI keyboards.

dpwe:
This integrates some more code from https://github.com/touchgadget/esp32-usb-host-demos to enable USB-MIDI keyboards (such as the M-Audio Oxygen49) to be connected to the USB-A port.  Received MIDI events are then forwarded to the Python handler.

The current MIDI host implementation does not support hot-swapping, so you can only plug in one USB device per boot.  To work with the MIDI keyboard, I have to use the idf.py monitor interface to type at the Tulip REPL.


bwhitman:
<3 
thank you dan! I will experiment with this and write it up

PR #99, opened 2023-06-25
amy: work-around esp32 bit-slip; fix large DC offsets in SAW waves

dpwe:
The int-range, soft-clipped sample values are further shifted down by 1 bit to be in range -16,384..16383 to avoid hard-wrapping on tulip/ESP32 board output.  (may be an issue with i2s bclk polarity, pending investigation).

alles.SAW_UP and SAW_DOWN waves exhibited large low-frequency bumps at their onsets, particularly with fade-in attacks.  This was because the DC offset, which is added to SAW waves to make their integration balance to zero, was constant across each block, while the waveform it was compensating was being linearly amplitude-scaled (when amplitude was varying).  By incorporating the DC offset into the render_lut routine, we can scale it by the linear amplitude ramp at the same time as the underlying looked-up value.  This is much cleaner and gets rid of the low-frequency excursions.

### DIFF
--- a/extmod/amy/src/amy.c
+++ b/extmod/amy/src/amy.c
@@ -822,7 +822,7 @@ int16_t * fill_audio_buffer_task() {
     //uint8_t nonzero = 0;
     for(int16_t i=0; i < BLOCK_SIZE; ++i) {
         for (int16_t c=0; c < NCHANS; ++c) {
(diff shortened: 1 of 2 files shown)


bwhitman:
this is great. thanks as always dan!!!!

PR #113, opened 2023-07-05
esp32s3/usb_keyboard.c: Support hot-swapping QWERTY<>MIDI.

dpwe:
This PR adds proper teardown for the USB Host when a QWERTY keyboard or MIDI keyboard is disconnected.  This makes the USB Host interface ready for another device to be connected. 

Before this change, either a QWERTY keyboard or a MIDI keyboard could be connected to the Tulip USB A socket, but once disconnected, the interface would not recognize any subsequent device connection.

After this change, the user can plug and unplug MIDI and QWERTY keyboards at will.  In particular, they can use a QWERTY keyboard to enter commands to launch software expecting MIDI input, then unplug the QWERTY keyboard, plug in a USB-MIDI keyboard, and play away.

In addition, this PR cleans up some of the diagnostic printouts that occur when DEBUG_USB define is set.

### DIFF
--- a/tulip/esp32s3/usb_keyboard.c
+++ b/tulip/esp32s3/usb_keyboard.c
@@ -19,73 +19,89 @@ typedef union {
     uint8_t val[9];
 } usb_hid_desc_t;
 
-#if DEBUG_USB
+//#define DEBUG_USB
+#ifdef DEBUG_USB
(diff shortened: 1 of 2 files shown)


bwhitman:
thank you dan!

PR #118, opened 2023-07-06
macos/virtualmidi: Bring MacOS MIDI in in-sync with esp32 FIFO changes.

dpwe:
This change at least allows it to compile Tulip Desktop on the Mac again.

I can't test it, but I expect it's OK.


bwhitman:
<3

PR #122, opened 2023-07-07
Update tulip_desktop.md

znmeb:
Ubuntu / Debian header packages end with `-dev`, not `-devel`. So it should be

`sudo apt install libsdl2-dev libffi-dev`


bwhitman:
Thanks!

PR #131, opened 2023-07-31
PCB v5

nvillar:
PCB V5R0 KiCAD project, including generated BOM and Schematic PDFs. Schematic ready for review; PCB not laid out or routed. Library complete with footprints, 3D models, datasheet links and order numbers for Mouser + DigiKey.


bwhitman:
Oh nice! Very excited to check this out. I'll merge it now and do some kicad learning this week to check it out. 
my only comment is I'd like to still call this "tulip4" for arcane and dumb reasons but we can change that later :)

PR #139, opened 2023-09-15
Alternative fix for missing touch release

nvillar:
Similar previous fix, but instead of periodically checking whether last touch was a TOUCH_EVT_PRESS, it checks whether a TOUCH_EVT_PRESS has been received since last forcibly sending a release event to MP.


bwhitman:
great, much better. thank you !

PR #148, opened 2023-12-07
T-Deck: Alt key implementation

hreikin:
This pull request gives the T-Deck an `alt` key toggle that works in the same way as the previously implemented `ctrl` toggle along with the logic to handle alt-keycodes. I have updated the header with defines for all the relevant `ints` used throughout the code and replaced them where required.

The KeyMappings structs `alternative` value now requires an `int` instead of `char` to avoid any unwanted conversion of numbers greater than 255. I have also updated the README with some tables to show all the implemented characters/shortcuts/keycodes.

While working on this PR i have uncovered a bug that i havent been able to track down yet and i dont think i have introduced. The alt-keycode functionality works well for the most part so if for example you pressed the required keys for `alt+36` it would display a `$` sign however if you press the required keys for `alt+163` it will not print the `£` sign. I think it stops working after 127 but i havent had time to debug this much further yet.

I did create a `boot.py` file with the following content to check if micropython could display the `$` or `£` signs. It displays the `$` sign without any issues but the `£` sign doesnt work correctly.

```
import tulip, time

print('$')
print('£')
print('$£')
time.sleep(5)
tulip.gpu_reset()
```

It seems to print a `:` and space before the `£` with the output looking like this on the T-Deck screen. 

```
$
: £
$: £
```
I havent gotten any further just yet but once i get time i will checkout your repo from before any of my changes and see if the `boot.py` file displays the same error or not. With all that being said id love to hear your thoughts on this so far ?


hreikin:
> . . . Clearly something we should fix; I will investigate that part. . . .

@bwhitman , have you managed to take a look into this at all ? i did try looking through the TFB/font related code but unfortunately i dont know C well enough to be able to debug this and offer any help

bwhitman:
Hey @hreikin ! I want to move the issue about non-lower ascii characters to a new issue so we give it more attention. Hold on a sec!

bwhitman:
We'll close this one for now and revisit the T-Deck keyboard if people are still actively using it -- we'll stay focused on the shipping Tulip board

PR #234, opened 2024-06-19
test commit i want to fail but be ok with a PR

bwhitman:
Testing a normal PR still works after protecting `main`


bwhitman:
this seems to work now. branch protected!

PR #255, opened 2024-07-13
Add a simple driver for the M5Stack 8encoder

mattytrentini:
Copied from: https://github.com/HowManyOliversAreThere/micropython-8encoder 

Device documentation available here: https://docs.m5stack.com/en/unit/8Encoder


bwhitman:
thank you !

PR #257, opened 2024-07-18
use XDG base directory spec for desktop builds if supported

stewlab:
This change adds the usage of the XDG Base Directory Specification when referencing `tulip.root_dir()` from Tulip Desktop. This includes updating the Linux build script to use `~/.local/share` instead of `~/Documents` when referencing the desktop version of the root directory.

XDG Base Directory Specification
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

**Testing**
I have tested this change in a Fedora 38 environment, and the output of `tulip.root_dir()` shows the new root directory in `~/.local/share/tulipcc` as expected.

I also built this from source and flashed to my Tulip CC device, and did not find any problems with functionality.  The output of `tulip.root_dir()` there was `/`.

**macOS**
I have not tested this for macOS. However, I noticed that version references `~/Documents` as well, so I provided fall back to that directory if `~/.local/share` does not exist. It seems that `~/Library` would be a better option instead of `~/Documents`, so it may be a good idea to update that as well later.

### DIFF
--- a/tulip/linux/build.sh
+++ b/tulip/linux/build.sh
@@ -8,10 +8,21 @@ make DEBUG=1
 mkdir -p dev
 cp build-standard/tulip/obj/tulip dev/
 
-mkdir -p ~/Documents/tulipcc
-mkdir -p ~/Documents/tulipcc/user
+tulip_home="${HOME}/.local/share/tulipcc"
 
-rm -rf ~/Documents/tulipcc/sys
-mkdir -p ~/Documents/tulipcc/sys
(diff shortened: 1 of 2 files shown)


bwhitman:
Thanks for this! I'll test it on macOS a bit and also ubuntu, debian and arch and WSL

bwhitman:
OK, this all works. I'll merge in the macOS fix and the doc fix next. Thank you for doing this!!!!

PR #310, opened 2024-08-13
Update tulip_desktop.md

znmeb:
Updated Linux build dependencies for Ubuntu / Debian and Fedora. I just ran these a few minutes ago. I can do openSUSE Tumbleweed too if you want, although I rarely use it.


bwhitman:
Thank you @znmeb !!

Issue #343, opened 2024-08-31
touchscreen scale should be settable at runtime


bwhitman:
A tulip user reached out to say their tulip's CTP was not responding where they thought it should. I built `world.download('paint.py')` for them to debug and it looks like the scale is off. We allow `touch_delta` to be set at runtime, but not `touch_scale` (which is set at 0.8 for the y-axis). We should allow scale to be a runtime var, and also update `calibrate` to divine a scale.

bwhitman:
https://github.com/user-attachments/assets/c38982cb-57ea-4756-a6ea-ae109f85ba8d

bwhitman:
We now ship with `y_scale` being settable by `touch_delta`. What we don't yet have is a version of `calibrate` that can guess it. I'll have some users try it out for me.

PR #345, opened 2024-09-03
Setting y-scale for touchscreen, #343

bwhitman:
Diagnostic tool for more control over CTP positioning. Try this with `world.download('paint.py')` 
(Also, upping AMY)


bwhitman:
#343

Issue #344, opened 2024-08-31
Tulip needs to understand MIDI "running status"


dpwe:
According to the MIDI description at http://midi.teragonaudio.com/tech/midispec/run.htm,
> The MIDI spec allows for a MIDI message to be sent without its Status byte (ie, just its data bytes are sent) as long as the previous, transmitted message had the same Status. This is referred to as **running status**.

This could explain the various reports of missing note-on/note-offs we've seen when receiving input from external MIDI gear.  This is specifically to reduce bandwidth and latency on serial-line MIDI, so it may be selectively employed with MIDI hardware but not pure software busses.

The effect is that "extra" data bytes without a Status byte (i.e., a byte with its top bit set) are implicitly assumed to have the same Status byte as the preceding command.  So a sequence like:
```
0x90  # Note-on Status, channel 0
0x3c  # Note 60
0x7f  # Velocity 127
0x90  # Note-on Status, channel 0
0x40  # Note 64
0x7f  # Velocity 127
0x90  # Note-on Status, channel 0
0x43  # Note 67
0x7f  # Velocity 127
```
can be sent equivalently without the repeated note-on Status bytes as:
```
0x90  # Note-on Status, channel 0
0x3c  # Note 60
0x7f  # Velocity 127
0x40  # Note 64
0x7f  # Velocity 127
0x43  # Note 67
0x7f  # Velocity 127
```

Currently, the Tulip midi parser doesn't understand this.

Many thanks to @Speccery for tracking this down!

Speccery:
Thanks for looking into this, I'll be looking forward to a release with the fix - or taking a look at the code myself.

bwhitman:
@Speccery please let us know if the latest Tulip (use `tulip.upgrade()` or Tulip Desktop build) works for you with your setup. We believe we've implemented support for running status and sense bytes.

AntoineDoinel69:
@bwhitman
I recently built a DIN to jack cable, and tried before and after the last update: it seems to work.
There are still problems intermixing notes with CC messages, but I'm not sure if its related.

bwhitman:
Hey @Speccery I've got my SH-4D wired up to Tulip over MIDI DIN, and sync is on MIDI and the sequences are playing in sync and with no dropped notes. Do you have a reduced test case we can try to show your error?

PR #355, opened 2024-09-06
Better midi parser, handles running status. Also support for all notes off

bwhitman:
- starting in on midi parser (and also side work on mpremote for it)
- fixed midi parser, and allowing midi.py to do all notes off


bwhitman:
I still can’t test running status but I’ve run through it in my head and it should work. All notes off (tested via Ableton) does work, and we now parse every midi message in the midi 1.0 specification properly (midi.py only handles a few, but it shouldn’t break other stuff) 

I tested USB midi on Tulip CC, still is fine

I tested IAC midi on Tulip Desktop mac, still is fine

bwhitman:
#344

PR #366, opened 2024-09-08
catching parsing errors for midi_cc_file.json

bwhitman:


bwhitman:
@dpwe can you take a look

dpwe:
Looks good to me. Those arp key bindings are very specific to my setup, we def don't want to depend on them.

dpwe:
I guess there's a similar problem in juno6.py - it will hit index errors with shorter midi control lists. But at least that doesn't run at startup.

PR #430, opened 2024-12-08
pointer fix for gt911_touchscreen.c

gunadai:
update gt911_touchscreen.c to set last_touch_x/y to -1 for all non-down pointers (currently hardcoded to 3)


bwhitman:
Great, thank you very much!

PR #466, opened 2025-01-31
Separate synth.py from midi.py

bwhitman:
douglas gave us good feedback about midi.py having all the Synth stuff in it was a bit confusing, so here's a first pass at separating out these. docs updated as well.


dpwe (approved):
Looks great!

I think we should probably call the Synth class `synth.PatchSynth()` - it's not actually a base class for all synthesizers, although the voice rotation stuff probably should be a base class.  What we currently call Synth is explicitly "a synthesizer constructed from voices that use the voices preset/stored patches mechanism".

I think in the future we'll have a VoicesSynth base-class that handles note-stealing, and PatchSynth will be a thin specialization of it, but that can come later.

I think the PCMSynth class that Douglas is kinda asking for will be a good future addition too.

dpwe (approved):
LGTM

PR #938, opened 2026-05-29
Detect out-of-date AMYboard firmware on editor sync

bwhitman:
## Summary

Makes the editor's **pull/sync** warn when a connected hardware AMYboard is running firmware older than the latest release. Related to #936 but a separate change.

- **Firmware build epoch** — `esp32_common.cmake` stamps the AMYboard firmware with `AMYBOARD_BUILD_EPOCH` (UTC unix seconds, seconds precision via CMake `TIMESTAMP "%s"`) and writes `build/amyboard_build_epoch.txt`.
- **SysEx query** — `modtulip.c` answers a new `zY` query with `F0 00 03 45 'Y' <ascii epoch> F7`. Handled in the tulip layer (guarded by `#ifdef AMYBOARD_BUILD_EPOCH`), so the **`amy` submodule is untouched** and older firmware just ignores the unknown opcode (no reply).
- **Version manifest** — `fs_create.py` emits `dist/amyboard-version.json` `{board, epoch, git_hash, build_date}`; `release.sh`/`dev.sh` upload it so `releases/latest/download/amyboard-version.json` is the comparison baseline.
- **Editor** — `spss.js` pre-warms the board epoch on the readiness ping, re-queries on sync, fetches the latest-release epoch through the existing `/api/firmware` proxy (always against `releases/latest`), and after a successful pull shows a dismissible `#firmwareOutdatedModal` that can jump to the Upgrade Firmware tab. A missing `zY` reply (pre-feature firmware) counts as out of date.

## Stacking note

This is **stacked on #936** (base branch `amyboard-dev-channel`), because the `dev.sh` edit here depends on `dev.sh`, which #936 creates. GitHub will auto-retarget this PR to `main` once #936 merges. Everything else (cmake, modtulip.c, fs_create.py, release.sh, spss.js, editor) is independent of #936.

## Test plan

- [x] AMYboard firmware builds with `-DAMYBOARD_BUILD_EPOCH` compiled in; `amyboard_build_epoch.txt` written.
- [x] `fs_create.py` writes `amyboard-version.json` with the matching epoch.
- [x] `spss.js` passes `node --check`; edits sync into `stage/`.
- [ ] **Hardware round-trip** (requires flashing a dev build): old release firmware on the board triggers the modal after a pull; a freshly flashed board does not.

### DIFF
--- a/tulip/amyboard/esp32_common.cmake
+++ b/tulip/amyboard/esp32_common.cmake
@@ -314,6 +314,15 @@ set(MICROPY_CROSS_FLAGS -march=xtensawin)
 # Suppress -Wgnu-folding-constant when building mpy-cross on macOS clang.
 set(ENV{CFLAGS_EXTRA} "$ENV{CFLAGS_EXTRA} -Wno-gnu-folding-constant")
 
+# Stamp the firmware with its build time (UTC unix epoch, seconds precision)
+# so the AMYboard editor can tell when a connected board is running firmware
+# older than the latest release. Recomputed on every CMake configure; both
+# release.sh and dev.sh clean-build (rm -rf build), so each published firmware
+# gets a fresh epoch. Also written to a file so fs_create.py can emit the
+# matching amyboard-version.json release asset with the same value.
+string(TIMESTAMP AMYBOARD_BUILD_EPOCH "%s" UTC)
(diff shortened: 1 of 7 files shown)


bwhitman:
this got re-built elsehwere

PR #994, opened 2026-06-16
amyboard: hardware-in-the-loop (HW CI) test prototype

bwhitman:
Prototype of a **hardware-in-the-loop test** for AMYboard (`tulip/amyboard/hwci/`): flash a built firmware to a connected board, reboot, drive it with USB-MIDI + AMY `zP` sysex, record the analog audio out, and RMS-compare to a committed reference for pass/fail — the hardware analogue of `amy/test.py`.

### Flow
1. **Firmware** from the per-PR preview (`amyboard-pr-<N>.vercel.app/firmware/amyboard-full-AMYBOARD.bin`); or `--firmware`/`--url`.
2. **Flash** via esptool over the serial debug header — RTS/DTR give auto download-mode + reboot (no buttons).
3. **Boot** → wait for the `AMYboard` USB-MIDI device.
4. **Drive**: timed USB-MIDI notes (C-major arpeggio on ch1, which the board sets up on boot via K257) + an AMY bleep via `zP` sysex.
5. **Record** the audio out from a USB audio input device (`sounddevice`).
6. **Compare**: onset-align + peak-normalize + RMS-diff in dB vs `ref/<name>.wav`, `--threshold` → PASS/FAIL.

### Status: Mac prototype, runnable in stages
`--list-devices`, `--flash-only`, `--no-flash`, `--update-reference`. Eventually a self-hosted runner on a **Raspberry Pi** runs this per-PR.

### Known next steps
- **PR audio upload**: GitHub's API can't attach binaries to PR *comments*; plan is to upload the recording as a **workflow artifact** + link it in the comment when this runs in CI (locally we save the wav).
- **Compare calibration**: hardware audio isn't sample-exact (latency/noise/levels/clock drift); the RMS threshold will need tuning on the bench once there's a stable reference.
- **CI wiring**: self-hosted-runner workflow on the Pi (label/dispatch-triggered).

Builds on the per-PR preview (#993) for firmware delivery.


dpwe:
I would love to build a cross-correlation compare function which should be very discriminant for checking the analog outputs. I'd need examples of the captured waveforms to develop. 

Might it be possible to capture bit-exact outputs via a spdif interface?

bwhitman:
yes, i'll ask you for a good audio audio comparison thing when we get there. great idea on s/pdif . i unfortunately don't have a s/pdif -> computer thing here. i may have this in a drawer: https://www.hifiberry.com/shop/boards/hifiberry-digi-io/  . if not i can get a s/pdif to USB thing.

PR #1006, opened 2026-06-19
amyboardweb: fix MIDI messages delivered 2x/4x in web editor

bwhitman:
## Summary

Dan reported that each MIDI CC from a LaunchControl was printed **2× (was 4×)** by a web sketch in the editor, while a native MIDI Monitor confirmed the controller sends each message **once**. The duplication is entirely in the web MIDI path (`tulip/amyboardweb/static/spss.js`) — two compounding bugs:

**1. Double delivery to Python (×2).** A non-sysex message reached MicroPython via *two* paths:
- **Path A:** `amy_process_single_midi_byte()` → AMY `convert_midi_bytes_to_messages` → `amy_event_midi_message_received` → main-thread `EM_ASM` → `amy_external_midi_input_js_hook` → `mp.midiInHook`.
- **Path B:** the explicit `mp.midiInHook` in the `"midimessage"` listener (added in e9cc3b09).

`amy_process_single_midi_byte` is cwrapped to the **main-thread** `am` module (`amyModule()`), so AMY's `EM_ASM` hook *does* reach `mp` — the "can't cross the AudioWorklet boundary" assumption in that commit was wrong. Both paths fired → every CC counted twice.

**2. Stacked listeners (×N).** `setup_midi_devices()` re-runs on the dropdown `onchange`, `statechange`, and the sync/pull flows. WebMidi.js `addListener()` *appends*, and the main `"midimessage"` handler was never removed — so each re-run stacked another copy. (The raw-sysex and passthru listeners were already removed before re-adding; the main one wasn't.)

Net effect was **(2 paths) × (N stacked listeners)** — hence 4×, then 2× after redundant `setup_midi_devices()` calls were reduced.

## Fix

- **Neutralize Path A:** `amy_external_midi_input_js_hook` is now a documented no-op, leaving the explicit listener call as the single source. This is the correct path to keep: AMY skips sysex in `convert_midi_bytes_to_messages` ("we do not pass sysex over to python"), so **sysex reaches Python only via Path B** — removing B instead would have silently broken sysex-to-Python.
- **Fix stacking:** `removeListener("midimessage")` before re-adding. Bonus: also fixes the double `move_knob` (CC-learn / live-knob) trigger that rode the same stacked listener.

Robust either way: even if Path A didn't fire, the no-op is harmless and the stacking fix still collapses N→1.

`tulip/web` is a **separate file** and intentionally left as-is — it relies on the AMY `EM_ASM` hook as its only path.

## Test

CI will spin up the PR preview site. To verify: open the editor on the preview, select a MIDI controller, load `midi_btns_patch_sel`, reveal the REPL, and confirm each press now prints **once**. (I could not drive Web MIDI headlessly here.)

### DIFF
--- a/tulip/amyboardweb/static/spss.js
+++ b/tulip/amyboardweb/static/spss.js
@@ -1128,9 +1128,17 @@ window.refresh_knobs_for_active_channel = async function(options) {
 };
 
 
-async function amy_external_midi_input_js_hook(bytes, len, sysex) {
-    mp.midiInHook(bytes, len, sysex);
-} 
+// AMY's main-thread WASM module calls this (via EM_ASM in
+// amy_event_midi_message_received) for every NON-sysex MIDI message it parses
+// out of amy_process_single_midi_byte(). We deliberately do NOT forward to
+// mp.midiInHook here: the "midimessage" listener in setup_midi_devices() below
+// already forwards every complete message (sysex included) to MicroPython
+// explicitly. Forwarding here as well delivered every non-sysex message to
+// Python twice — the "MIDI CC repeated 2x" bug. Kept as a no-op (not removed)
+// so AMY's `typeof amy_external_midi_input_js_hook === 'function'` check still
+// passes. NOTE: tulip/web (not this file) still relies on this hook as its only
+// path, so don't "fix" it the same way there.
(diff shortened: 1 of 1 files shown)


bwhitman:
dan confiremd this fixes it

PR #1015, opened 2026-06-20
amyboardweb editor: synth teardown on disable + patch restore on re-enable; drop dead .patch-file code

bwhitman:
## What & why

Addresses Dan's note that the editor's **Active** checkbox "doesn't do much." Previously: disabling a channel was a no-op client-side flag (the synth kept playing), and re-enabling hard-reset the channel to the default **K257** patch — which is the "resets the patch to 257" he saw.

### Active checkbox now controls the synth (gated to the user toggle only)

- **Disable** → `num_voices=0` for that channel. In AMY this calls `instrument_release()` + `midi_clear_channel_mappings()`, so the channel's voice oscillators are freed, its CC maps cleared, and it goes silent / stops using polyphony.
- **Re-enable** → rebuild the voice alloc (`K257` + 6 voices), re-send the CC mappings (required since teardown cleared them), then **re-apply the channel's existing knob values** (new `send_all_knob_values()` helper) plus the dedicated **Level** slider — so the channel comes back sounding the way it did, instead of snapping to the bare default patch.

Both behaviors are behind a `userToggle` option, so **only the checkbox** triggers them; startup and sketch/dump-load callers keep their existing behavior.

This matches Dan's sketch (`num_voices=0` to tear down; on re-enable set 6 voices, reload K257, reuse existing knobs) — and goes one step further by restoring the prior knob **values**, not just the CC mappings.

### Also: remove dead `.patch`-file code

While wiring restore I found it was about to lean on the old on-disk `.patch` feature, which was dropped months ago — the UI elements are gone, only `getElementById(...)→null` no-op handlers remained. Removed:

- `load_saved_patch_file_into_current_channel` / `save_current_synth_patch_file` / `list_current_patch_files`
- the "Your Patches" / Save Patch / Revert handlers
- the orphaned World **"Patches"** sub-tab (browse/search/preview/download/upload) and all its JS — it was entirely `.patch`-file based and half-broken

**Untouched / still live:** the Load Preset modal (Juno/DX7 presets via K-numbers), the World **"Sketches"** tab, and the per-channel Clear button.

Net: **+58 / −821** across `spss.js` and `editor/index.html`.

## Testing

Please test on the PR preview (`amyboard-pr-<N>.vercel.app`) → editor:
- Toggle **Active** off on a channel that's playing → it should stop.
- Tweak some knobs + Level, toggle Active off then on → the sound (and Level) should come back as you left it, **not** reset to the default patch.
- Confirm Load Preset (Juno/DX7) and the World → Sketches tab still work.

Local: `node --check` passes on all inline scripts and `spss.js`; submodules bootstrapped and a local `dev.py` build/serve was run for a load smoke-test.

### DIFF
--- a/tulip/amyboardweb/static/editor/index.html
+++ b/tulip/amyboardweb/static/editor/index.html
@@ -1116,19 +1116,6 @@
         }
       }
 
-      // Load patches on first visit to the Patches sub-tab.
-      var worldPatchTab = document.getElementById("world-patch-tab");
-      if (worldPatchTab) {
-        worldPatchTab.addEventListener("shown.bs.tab", function() {
-          if (typeof refresh_amyboard_world_patches === "function") {
-            refresh_amyboard_world_patches();
-          }
-          if (typeof refresh_patch_upload_dropdown === "function") {
-            refresh_patch_upload_dropdown();
-          }
-        });
-      }
-
       // When the main World tab is shown, refresh the patch dropdown if
       // the Patches sub-tab is currently active (Bootstrap won't re-fire
       // shown.bs.tab on the sub-tab in that case).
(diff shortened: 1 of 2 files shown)


dpwe:
I've tested this on an AMYboard.  I confirm that toggling the Active checkbox immediately silences the corresponding channel (while leaving other channels unmodified).  Re-enabling Active restores a K257 patch with current knob settings to the synth.

When a DX7 synth is de-activated, and the channel is re-activated, we get a K257 (analog) patch, not the previous DX7 patch.  This is not ideal, but it's not a regression - I believe it did this before also, but it didn't silence the synth when inactive.  Maybe something we could fix in the future.

PR #1022, opened 2026-06-21
web: serve COOP/COEP cross-origin isolation headers (fix audio in iOS WKWebView / MIDIWeb)

bwhitman:
## Problem

Per [5of12/MIDIWeb-Hub#23](https://github.com/5of12/MIDIWeb-Hub/issues/23): Tulip Web (`/run`) and the AMYboard editor (`/editor`) produce **no audio inside iOS WKWebView** (e.g. the MIDIWeb app), even though they work in iOS Safari.

Root cause is cross-origin isolation. Both apps need `crossOriginIsolated === true` (for `SharedArrayBuffer` + AudioWorklet + WASM), which requires the document to be served with:

- `Cross-Origin-Opener-Policy: same-origin`
- `Cross-Origin-Embedder-Policy: require-corp`

Today those headers are injected client-side by the `mini-coi.js` **service-worker shim**. Safari honors it; **iOS WKWebView does not**, so isolation never happens and audio is silent.

## Fix

Serve the COOP/COEP/CORP headers **directly from Vercel** via `vercel.json`, so isolation no longer depends on a service worker.

**Scoped to the app routes** (`/run`, `/editor`) rather than `/(.*)`. The marketing landing pages embed a cross-origin **YouTube iframe** and a **createsend newsletter script** — under a global COEP `require-corp`, the newsletter script (no CORP header) would be blocked. Tulip's landing page has no COEP today, so a global rule would be a regression. Scoping targets exactly the pages that need isolation and leaves the marketing pages untouched.

Verified the app routes are COEP-safe: every cross-origin CDN dependency (codemirror, bootstrap, webmidi, jquery from cdnjs / jsdelivr / googleapis) already returns `Cross-Origin-Resource-Policy: cross-origin`, and all same-origin WASM/JS passes COEP inherently. This is the same COEP policy the apps already run under in Safari via `mini-coi`, so resource loading is unchanged.

`mini-coi.js` is left in place as a harmless fallback for non-Vercel/local serving (it sets identical headers).

### Also: `tulip/web/vercel.json` was never deployed

The release/preview workflows run `vercel deploy stage`, but `tulip/web/build.sh` never copied `vercel.json` into `stage/` — so the file was silently ignored in production (its existing cache rules weren't applied either; hashed assets were served with Vercel's default `max-age=0` instead of `immutable`). Added the `cp vercel.json stage/` that `amyboardweb/dev.py` already does.

## How to test (preview deploys)

- **Tulip Web:** `https://tulip-pr-<N>.vercel.app/run/`
- **AMYboard:** `https://amyboard-pr-<N>.vercel.app/editor/`

On each, in the console:
```js
window.crossOriginIsolated === true
typeof SharedArrayBuffer === "function"
```
…and confirm audio (Tulip: bleep on first interaction; AMYboard: Simulate → tap a key). Marketing landing pages (`/`) should be unchanged (no COEP header; YouTube embed still loads).

Final verification of the actual MIDIWeb/WKWebView case needs an iOS device.

### DIFF
--- a/tulip/amyboardweb/static/mini-coi.js
+++ b/tulip/amyboardweb/static/mini-coi.js
@@ -2,6 +2,12 @@
 /*! mini-coi - Andrea Giammarchi and contributors, licensed under MIT */
 (({ document: d, navigator: { serviceWorker: s } }) => {
   if (d) {
+    // If the server already sends COOP/COEP (Vercel prod/preview + webdev.py),
+    // the page is cross-origin isolated without us. Registering the SW then only
+    // adds a harmful first-load location.reload() that can interrupt the WASM
+    // boot (observed as a blank app / stray textarea on iOS Safari & WKWebView).
+    // Only fall back to the service-worker shim on header-less static hosts.
+    if (self.crossOriginIsolated) return;
     const { currentScript: c } = d;
     s.register(c.src, { scope: c.getAttribute('scope') || '.' }).then(r => {
       r.addEventListener('updatefound', () => location.reload());
--- a/tulip/amyboardweb/vercel.json
+++ b/tulip/amyboardweb/vercel.json
@@ -7,6 +7,22 @@
   ],
   "headers": [
     {
+      "source": "/editor",
+      "headers": [
+        { "key": "Cross-Origin-Opener-Policy", "value": "same-origin" },
+        { "key": "Cross-Origin-Embedder-Policy", "value": "require-corp" },
+        { "key": "Cross-Origin-Resource-Policy", "value": "cross-origin" }
+      ]
+    },
+    {
+      "source": "/editor/(.*)",
+      "headers": [
+        { "key": "Cross-Origin-Opener-Policy", "value": "same-origin" },
+        { "key": "Cross-Origin-Embedder-Policy", "value": "require-corp" },
+        { "key": "Cross-Origin-Resource-Policy", "value": "cross-origin" }
+      ]
+    },
+    {
       "source": "/((?!amy-|amyboard-).*)",
       "headers": [
         {
--- a/tulip/web/build.sh
+++ b/tulip/web/build.sh
@@ -43,6 +43,12 @@ cp site/index.html stage/
 cp -Rf site/webfonts stage/
 cp -Rf ../../assets/css ../../assets/fonts ../../assets/img ../../assets/js stage/
 
+# vercel.json must sit at the deploy root: the release/preview workflows run
+# `vercel deploy stage`, so a vercel.json left in tulip/web/ is ignored. Copy it
+# in so its headers (COOP/COEP cross-origin isolation for /run, plus cache rules)
+# actually get served. (amyboardweb's dev.py does the equivalent copy.)
+cp vercel.json stage/
+
 cp ../../amy/build/amy.js stage/run/amy-$timestamp.js
 cp ../../amy/build/amy.wasm stage/run/amy-$timestamp.wasm
 cp ../../amy/docs/amy.aw.js stage/run/amy-$timestamp.aw.js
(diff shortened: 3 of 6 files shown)


bwhitman:
## Verified on the Vercel previews ✅

Both previews build & deploy cleanly, serve the isolation headers on the **app routes only**, and the apps come up cross-origin isolated with **no COEP-blocked resources**. Checked with `curl` (headers) + headless Chrome via CDP (`crossOriginIsolated` / `SharedArrayBuffer` / blocked-resource + console scan).

| route | COOP/COEP/CORP | `crossOriginIsolated` | `SharedArrayBuffer` |
|---|---|---|---|
| `tulip-pr-1022.vercel.app/run/` | ✅ all three | `true` | `function`, constructs OK |
| `tulip-pr-1022.vercel.app/` (marketing) | none | `false` | — |
| `amyboard-pr-1022.vercel.app/editor/` | ✅ all three | `true` | `function`, constructs OK |
| `amyboard-pr-1022.vercel.app/` (marketing) | none | `false` | — |

- **No COEP-blocked resources** on either app route — every cross-origin CDN dep (codemirror / bootstrap / webmidi / jquery) already sends `Cross-Origin-Resource-Policy: cross-origin`.
- `/run/` runtime behaviour is **identical to current production** (same single pre-existing `memory access out of bounds` headless-only main-loop error + benign `ERR_ABORTED`; **no new errors**). Confirmed by running the same check against `tulip.computer/run/`.
- The `tulip/web/vercel.json` cache rules now actually apply too (hashed assets → `immutable`) — the file was previously never deployed.

**Remaining:** on-device verification in MIDIWeb (iOS WKWebView), which CI can't do. The preview URLs are already isolated, so loading `https://amyboard-pr-1022.vercel.app/editor/` (Simulate → tap a key) or `https://tulip-pr-1022.vercel.app/run/` (bleep on first interaction) **in MIDIWeb** should now produce audio without merging to prod.

bwhitman:
## ✅ Confirmed working on a real iOS device

### Two distinct bugs were in play

1. **iOS WKWebView / MIDIWeb had no audio** — the original report. Cause: cross-origin isolation came only from the `mini-coi.js` service-worker shim, which iOS WKWebView doesn't honor → no `SharedArrayBuffer`/AudioWorklet. **Fix:** serve `COOP`/`COEP`/`CORP` from the server (scoped to `/run` and `/editor`).

2. **`/run` was blank on iOS Safari** — surfaced while testing. Cause was *not* COOP/COEP: Tulip's `/run` page references its assets **relatively**, so loading `/run` **without a trailing slash** resolved them against `/` and 404'd everything (`loadMicroPython`/`start_audio` undefined → blank page + stray textarea). Pre-existing on prod too. **Fix:** `<base href="/run/">`, mirroring amyboardweb/editor's existing `<base href="/">` (which is why the editor already worked at both `/editor` and `/editor/`). No redirect needed.

Also: `tulip/web/build.sh` now copies `vercel.json` into `stage/` (it was never deployed before, so its cache rules were dead too), fixed an invalid Vercel route pattern (nested capture group), and made `mini-coi.js` bail when already isolated (redundant now that the server sends the headers).

### Verification (headless Chrome on the previews)

| route | crossOriginIsolated | SharedArrayBuffer | 404s | COEP-blocked |
|---|---|---|---|---|
| `tulip-pr-1022…/run` (no slash) | ✅ true | function | 0 | 0 |
| `tulip-pr-1022…/run/` | ✅ true | function | 0 | 0 |
| `amyboard-pr-1022…/editor` | ✅ true | function | 0 | 0 |
| `amyboard-pr-1022…/editor/` | ✅ true | function | 0 | 0 |

Marketing landing pages (`/`) remain un-isolated (COEP scoped to app routes only, so their YouTube/newsletter embeds are unaffected). Confirmed on a physical iOS device.

PR #1042, opened 2026-06-25
web: bridge get_synth_commands() to JS callers in both web apps

bwhitman:
## What

On web, AMY runs in a separate WASM worklet (`AMY_IS_EXTERNAL`) and the micropython module does **not** link it — so `tulip.amy_get_synth_commands` (`modtulip.c`) sits inside `#ifndef __EMSCRIPTEN__` and is compiled out on web. The convenience `get_synth_commands` only existed for "micropython that links AMY" (desktop/ESP) and for CPython (`amy/src/pyamy.c`).

This bridges the same capability to **JS callers** on web by driving AMY's already-exported low-level generator `yield_synth_commands`. Pure JS in `static/` — no AMY/submodule change.

## Changes

- **`tulip/amyboardweb/static/spss.js`** — adds a canonical global `get_synth_commands(synth, include_fx = true)` returning an array of wirecode strings (JS analogue of pyamy.c's wrapper). The editor-only `get_wire_commands_for_channel` now delegates to it (same error contract; heap-buffer free moved into `try/finally`).
- **`tulip/web/static/spss.js`** — Tulip Web lacked the plumbing, so adds the `yield_synth_commands` cwrap, a `read_c_string_from_heap` helper, and the identical `get_synth_commands`.

Both files are plain `<script>`s, so `get_synth_commands` is a true global (also assigned to `globalThis`).

```js
const cmds  = get_synth_commands(5);        // wirecode strings for synth 5
const noFx  = get_synth_commands(5, false); // omit global FX commands
```

## Verification

- `node --check` passes on both files.
- Isolated logic test with a mock AMY module: generator loop terminates correctly, empty synth → `[]` (matches pyamy/modtulip), `include_fx` passes through, range validation (0 / 17 / 2.5) throws, and malloc/free stay balanced across all paths including throws.
- Reuses the exact cwrap+loop already running in amyboardweb; AMY's `make web` exports `_yield_synth_commands` unconditionally.

Not done: full WASM build / live browser run (static-JS-only change over an already-proven path).

### DIFF
--- a/tulip/amyboardweb/static/spss.js
+++ b/tulip/amyboardweb/static/spss.js
@@ -520,35 +520,62 @@ function get_wire_commands_for_juno_patch(patch) {
   return wire_commands;
 }
 
+// Canonical JS bridge for AMY's get_synth_commands. The C convenience wrapper
+// exists for CPython (amy/src/pyamy.c) and for MicroPython as
+// tulip.amy_get_synth_commands (modtulip.c) -- but the MicroPython one is compiled
+// out on web (#ifndef __EMSCRIPTEN__) because here micropython does not link AMY;
+// AMY runs in a separate WASM worklet. So we bridge it from JS instead, driving
+// AMY's exported low-level generator yield_synth_commands. Reads back the wirecode
+// commands that reconstruct synth `synth` (1..16) from AMY's current state and
+// returns them as an array of strings (empty if the synth has no state).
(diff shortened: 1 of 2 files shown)


bwhitman:
Folded into #1041 — cherry-picked commit `8778363` onto that branch (as `d5b66aab`), since it's closely related to the SYNC 2 work there. Closing in favor of #1041.

PR #1044, opened 2026-06-26
amyboard: runtime MIDI OUT TRS Type A/B switch

bwhitman:
## What

Adds a runtime way to switch the AMYboard MIDI **OUT** between TRS **Type A** and **Type B**, from Python, without a reboot and without restarting AMY:

```python
import amyboard
amyboard.set_midi_type('B')   # switch MIDI OUT to Type B
amyboard.set_midi_type('A')   # back to Type A (the boot default)
amyboard.midi_type()          # -> 'A' or 'B', whichever is live
```

Not persisted — AMYboard always boots as Type A (no NVS).

## Why

The MIDI OUT 3.5mm TRS jack has two incompatible wiring standards: **Type A** (the MIDI-spec default, most modern gear) and **Type B** (older Arturia/Novation/Akai, etc.). They swap the TRS tip/ring. Previously `init_midi()` set this once at boot (default A) with no way to change it live — if your external device expected the other type, you had no recourse short of editing firmware.

## How

On AMYboard, "Type A vs B" is simply which GPIO carries MIDI OUT: **Type A = pin 14, Type B = pin 15** (the unused leg is held high at MIDI idle). AMY transmits MIDI via `uart_write_bytes(UART_NUM_1, ...)` — keyed on the **UART number, not the pin** — so switching type is just re-pointing that UART's TX line with `uart_set_pin()`. No `amy_start()` re-run (which would have leaked `external_map` and restarted audio). Only MIDI OUT is affected; MIDI IN (pin 21) works with either type and is untouched.

## Changes

- `tulip/shared/amy_connector.c` — `amyboard_set_midi_out(pin)` (AMYBOARD-only): `uart_set_pin()` on `UART_NUM_1` + hold the now-unused TRS leg high.
- `tulip/shared/modtulip.c` — `tulip.amyboard_set_midi_out(pin)` MicroPython binding.
- `tulip/shared/amyboard-py/amyboard.py` — `set_midi_type('A'|'B')` + `midi_type()`; `init_midi()` refactored to share one pin map. Boot path (`start_amy` → `init_midi()`, default A) unchanged.
- `docs/amyboard/python.md` — documented under "Sending MIDI out".

No AMY submodule changes.

## Verification

- ✅ Compiles + links into AMYboard firmware — full `idf.py -DMICROPY_BOARD=AMYBOARD build`, exit 0, no warnings on touched files. Built against the pinned AMY commit `ec78cc8b`, where `midi_uart=1` → `UART_NUM_1` is confirmed.
- ⏳ Hardware electrical verification (OUT signal actually moves 14↔15, both directions drive a real MIDI device) — pending HW CI / bench.

### DIFF
--- a/docs/amyboard/python.md
+++ b/docs/amyboard/python.md
@@ -262,6 +262,22 @@ amy.send(osc=0, note=60, vel=0, sequence="24,48,2")   # note off at tick 24 of e
 
 `amy.AMY_MIDI` always sends on MIDI channel 1, and notes that arrived over MIDI in are not echoed back out. See the [AMY MIDI docs](https://github.com/shorepine/amy/blob/main/docs/midi.md#sending-midi-out) for the full details.
 
+#### MIDI out TRS type (A vs B)
+
+The MIDI out jack is a 3.5mm TRS connector, and there are two incompatible wiring standards for these: **Type A** (the official MIDI-spec standard, used by most modern gear) and **Type B** (older, used by some early Arturia/Novation/Akai devices). They swap the TRS tip and ring, so if your external device isn't receiving MIDI, it likely expects the other type. AMYboard defaults to **Type A**; only MIDI out is affected — MIDI in works with either type.
+
+Switch it at runtime with `amyboard.set_midi_type()` (no reboot needed; AMY keeps running):
+
+```python
+import amyboard
+
+amyboard.set_midi_type('B')   # switch MIDI out to Type B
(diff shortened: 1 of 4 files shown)


bwhitman:
I don't have any MIDI type B equipment or cables. but i do know that this works with a type a cable with 'A' selected and does not work with 'B' selected. so at the worst this does no harm

PR #1050, opened 2026-06-28
Update m5_8encoder.py to allow for correct addressing of the 9th LED

AraKnine:
set_led() was checking to see if the value of 8 was less than the number of ENCODERS(8) instead of the number of LEDS(9). This update will have it check against LEDS(9) to allow for the LED beside the SW switch to be addressed correctly.


bwhitman:
Thanks!!!

PR #1116, opened 2026-07-09
Port pye to Tulip: full TFB ANSI support, editor app with colors/shortcuts, grab_tty, GC-root fix

bwhitman:
Makes [pye](https://github.com/robert-hh/Micropython-Editor) — the editor AMYboard has used all along — Tulip's default editor, running on the TFB with the old C editor's look and feel, and fixes several long-standing bugs found along the way.

## TFB ANSI/VT100 support (display.c)

- Rewrote escape parsing as a stateful machine: sequences split across stdout writes parse correctly, unknown non-CSI escapes (`ESC M`) no longer leak bytes onto the screen, and the old out-of-bounds `str[i+1]` read is gone
- `ESC[{row};{col}H` is now 1-indexed like a real terminal (was 0-based, shifting everything down-right a cell); added `ESC[{n}A/B/C/D` relative moves and the `f` alias
- `ESC[?25h/l` cursor show/hide
- `ESC[K` modes 0/1/2 and `ESC[J` modes 0/1 (previously nyi), erasing with the active colors ("background color erase") like real terminals
- 24-bit SGR (`ESC[38;2;r;g;bm` / `48;2`) mapped exactly onto the 332 pal; raised the SGR param cap from 5 to 8 (the old clamp silently dropped the blue component)
- Fixed SGR attribute-off codes 24–29 (`=-` typos corrupting format state)
- The cursor no longer stamps default colors onto cells it visits (this also fixes the old REPL quirk of the cursor bleaching colored text)

## pye as the Tulip editor

- `edit()` / `tulip.edit()` open pye; the old C editor remains as `old_edit()`. Tulip Web keeps the old editor (pye needs blocking key reads)
- Same colors and syntax highlighting as the old editor (strings/comments/numbers/operators, exact pal colors via 24-bit SGR), inverse status bar, selection bg
- Both keymaps: pye's own (ctrl-S/F/Z/G/L/C/O/R, ESC-ESC quit) plus the old editor's (ctrl-X save, ctrl-W search, ctrl-K/U cut/paste, ctrl-A/E, ctrl-Y/V paging)
- Runs as a UIScreen app: quit + alt-tab buttons, ctrl-Q/ctrl-tab work, and the REPL keeps running while the editor is open (pye's loop runs on a `_thread` fed by `keyboard_callback`)
- `grab_tty`: over serial (or Desktop in a terminal), typing drives the editor, AMYboard-style — `tulip.tty_grab()`/`tulip.tty_read()` divert console input from the REPL into a ringbuffer; any UIScreen can opt in with `grab_tty=True`
- Physical PgUp/PgDn/Home/End get their own key codes translated to proper ANSI (they used to collide with ctrl-Y/ctrl-V, so PgUp in pye would have been *redo*); the C editor gains Home/End

## Crash fixes

- **All C-held Python callbacks are now GC roots** (`MP_REGISTER_ROOT_POINTER`): `ui_quit_callback`, `keyboard_callback`, `frame_callback`, defer/sequencer slots, midi/amy callbacks. A bound method whose only reference was a plain C global got collected, and the next ctrl-Q crashed in `mp_sched_run_pending` — latent on every port for years, made near-deterministic by pye's allocation churn. Any historical mystery crash with that stack was probably this
- `tulip.tfb_restore()` without a prior save was a NULL deref; now a guarded no-op, and double-restore is safe

## Verification

Driven end-to-end on Tulip Desktop via scripted stdin: TFB cell readback asserts for cursor addressing, split sequences, SGR/truecolor, bce, syntax colors; full pye sessions (open/type/save/quit); the real ctrl-Q-through-scheduler crash path with forced GC; alt-tab away/back with screen restore; and raw-TTY editing including escape-sequence arrows. Desktop, TULIP4_R11, AMYBOARD and Tulip Web builds all pass.


bwhitman:
This mostly works on tulip hardware but is .... slower ... than my C one. I think i may go back and make the C one better, but TBD.

bwhitman:
We're not going to do it this way

PR #1132, opened 2026-07-10
AMYboard Web: per-bus effects UI (buses from amy#686)

bwhitman:
Implements dpwe's "Adding buses to the AMYboard Web Editor" request (2026-07-10).

## What's new

- **Bus pull-down in the channel strip** (next to Level): assigns the selected channel's synth to mix bus 0–3, sending `amy.send(synth=<ch>, bus=<bus>)` (`i<ch>y<bus>`). Default 0.
- **Per-bus Effects column**: the EQ/Chorus/Reverb/Echo knobs now mirror state **per bus** (`amy_bus_knobs[0..3]`), not globally. Changing the bus pull-down (or switching to a channel on another bus) swaps the whole Effects column to that bus's knob positions. Moving FX knobs on channel 2 (bus 1) shows the same positions when you switch to any other channel on bus 1.
- **Master Level in the Effects panel** (parallels the channel Level slider): sets the bus's mixdown volume (`y<bus>V<val>`). Clicking its label opens the same MIDI-CC popup as other knobs (template `i%iV%v`). Placed at the top of the Effects panel rather than the desktop-only header bar so it also works on mobile, where that bar is hidden.

## Design decisions (flagging for @dpwe)

1. **FX knob wire templates are synth-prefixed** (`i%ix,,%v` etc.), per the "neater and safer" suggestion — both live knob moves and device-side MIDI-CC mappings resolve the bus through the channel at execution time, so CC maps survive bus reassignment.
2. **The `_auto_generated_knobs` log records FX bus-explicit** (`y0x,,-3.059`) instead of synth-prefixed. The log collapses last-wins by structural key, so a synth-keyed FX slot can only hold ONE bus's history — move the channel to another bus, tweak the same knob, and the old bus's value would be overwritten *in place*, replaying against the wrong bus on boot. One `y<bus>` slot per bus per parameter avoids that entirely, and is order-independent w.r.t. bus assignments. Legacy unprefixed FX lines load as bus 0.
3. **AMY pin bump to latest main (79f1337) is part of this PR** — required, not incidental: the readback FX line only gained `y<bus>` and scalar `volume` only became bus-directed (`i2V0.4` → synth 2's bus volume, `play_delta` bus offset) after the current pin (3a7c176 / v1.2.43). The bump is small (bus fixes, `instrument_get_bus` return type, sysclock integer-math fix). `tulip/server/refdocs/amy` is byte-identical after `sync_amy_docs.py`; the `*.generated.js` files are unchanged (no `__init__.py`/`patches.h` changes).
4. `reset_global_effects()` is per-bus and now also **drops that bus's FX lines from the knob log**, so Clear/preset-load leaves the sketch consistent with the reset state (previously stale global FX lines survived in the block).
5. Changing the bus on an **inactive** channel updates the UI mirror only — no `i<ch>y` line is recorded, since AMY ignores bus assignments for nonexistent synths and a stray line would make the channel read as active on the next load.

## Known limitations

- AMY's built-in K257 default CC map still carries unprefixed FX templates (`k%v`, `h%v`, `M%v`), so hardware CC through the *default* map targets bus 0 until the editor pushes its own mappings. Fixing that is an amy `patches.h` change (follow-up).
- Old sketches whose blocks contain the old unprefixed FX CC templates (`ic93,...,k%v`) won't match the new knob templates, so those CC numbers won't show in the UI (they still work on-device via block replay, targeting bus 0).
- Hardware running firmware older than this amy pin will route scalar `V` to bus 0 regardless of prefix; firmware ships from the same pin on merge, so this only affects not-yet-upgraded boards in control mode.

## Tested (web simulator, `dev.py`)

- FX knob move on ch1/bus0 → live `i1x,,-3.059`, log `y0x,,-3.059Z`; AMY readback (`get_synth_commands`) confirms.
- Bus switch → `i1y1Z` recorded (collapses in place on re-switch), Effects column + master Level + "Bus N" indicator swap to the new bus's mirror.
- Chorus move on bus 1 → `y1k0.500Z`; readback FX line `y1V1.000x...k0.500...` confirms AMY state.
- Master Level on bus 1 → `y1V0.400Z`, `amy_bus_volumes[1]=0.4`.
- Full position-from-sketch (page load / Read): channel buses, per-bus FX, bus volume, select/indicator all restored from the block alone; localStorage draft round-trips across reload.
- Synth readback after boot no longer resets other buses' mirrors (onlyDirtyFxBuses).
- Legacy pre-bus block (`x,,-5`, `h0.3`, `M0.8`) positions to bus 0.
- Clear channel: synth back to K257/bus 0, bus FX lines pruned from the log.


dpwe:
Nice!

The bus level slider would ideally fit in the "effects" column header (text could be shortened to "FX"). I honestly don't think we need the value box (for *either* level slider), and we definitely don't need the "Bus 1" text - that's redundant with the Bus pulldown in the channel header

dpwe:
This looks nice.  Can we change both level sliders to have a pseudo-logarithmic range, with 1.0 as "mid slider", something like 10.0 at the right edge, and something like 0.0 at the left edge?  For instance, if the slider position is 0 <= p <= 1, the actual level could be (pow(10, 2 * (p - 0.5)) - 0.1)

PR #1139, opened 2026-07-11
User C DSP plugins: design doc + amy bus-dsp-hook pin (amy#874)

bwhitman:
First slice of the user-C-DSP-plugins idea: let people write a small C effect or oscillator as a Python string, compiled **on-device** (xcc700 → Xtensa REL ELF → esp-idf elf_loader on the S3) and injected into AMY's render loop.

This PR contains:
- `docs/user_c_dsp_design.md` — full design & feasibility write-up. Core unknowns were validated by experiment: xcc700 compiled a bit-crusher to a 608-byte Xtensa ELF with a clean windowed-ABI disassembly and an exported `process` symbol; elf_loader runs loaded code from PSRAM on S3 and resolves imports against a firmware symbol table; AMY's int32 S8.23 `SAMPLE` type matches the compiler's int-only subset.
- amy submodule pinned to shorepine/amy#874 (`amy_external_bus_dsp_hook` — the one AMY change needed: an effect hook at the end of each bus' FX chain). The oscillator side needs no AMY change (`amy_external_render_hook` already exists and is chained in `amy_connector.c`).

Testing:
- amy: `make test` — 109 tests pass bit-exact at the pinned commit.
- tulipcc: Tulip Desktop macOS builds against the pin; headless stdin smoke test boots, `amy.send(osc=0, wave=SINE, freq=440, vel=1)` accepted.

**Do not merge until amy#874 merges** — then the pin moves to the merged amy commit (and `tulip/server/sync_amy_docs.py` refresh rides along, since `docs/api.md` changed).

Next slices per the design doc: vendor xcc700 + elf_loader component into the amyboard firmware, `tulip.install_c_process()` binding, effect dispatch in `amy_connector.c`, bench test on hardware.


bwhitman:
Pushed the desktop implementation (a9a0b804): `tulip.install_c_process()` is live on Tulip Desktop via an embedded libtcc JIT (new `tinycc` submodule) — no Xcode CLT needed. Dispatcher registered as `amy_external_bus_dsp_hook` (amy#874), user contract `void process(int *buf, int frames, int chans)` over int32 s8.23 SAMPLE buffers.

Try it after `./build.sh` in `tulip/macos`:

```python
src = '''
enum { BITS = 18 };
void process(int *buf, int frames, int chans) {
    int i = 0;
    while (i < frames * chans) {
        buf[i] = (buf[i] >> BITS) << BITS;
        i = i + 1;
    }
}
'''
tulip.install_c_process('crush', src)
tulip.c_process('crush', True)        # on bus 0; (name, on, bus=0)
amy.send(osc=0, wave=amy.SINE, freq=220, vel=1)
tulip.c_process_calls('crush')        # ~86/sec if it's running
tulip.c_process('crush', False)
```

Bad C raises `ValueError` with tcc's diagnostics; reinstalling the same name hot-swaps the code and keeps bus enables. `package.sh`/CI builds are untouched (feature compiles out without `tinycc/libtcc.a`; signed builds will need the allow-jit entitlement + per-arch tcc later). Verified with a headless stdin test: JIT'd crusher runs per-block on the render thread, uninstall mid-playback is clean.

bwhitman:
a8d12964: applied dpwe's amy#874 review downstream (re-pinned to `8d0f470`, dispatcher renamed to `tulip_bus_postprocess_hook`) — and **breaking change to the user contract**: effects now see plain 16-bit PCM instead of s8.23 fixed point, so authors don't need to know AMY's internal format.

```c
#include <stdint.h>
void process(int16_t *buf, int frames, int chans);  // samples are -32767..32767, like a WAV
```

The dispatcher converts (shift+clamp per sample) at the boundary, only on buses that have an effect enabled. Any earlier `int *buf` snippets need their types + shift counts updated (e.g. bitcrusher BITS 18 → 10).

bwhitman:
3c1d8195 adds the two ergonomics asks:

**Body-only snippets** — send just the loop; the wrapper adds the signature + `stdint/stdio/math` includes (`#line` keeps tcc error line numbers pointing at your code):

```python
tulip.install_c_process('crush', '''
    int i = 0;
    while (i < frames * chans) {
        buf[i] = (buf[i] >> 10) << 10;
        i = i + 1;
    }
''')
```

**User C oscillators** — `install_c_osc(name, src)` + `c_osc(name, osc)` replace an AMY osc's waveform with your C. The body gets `buf` (mono int16), `frames`, `osc`, `phase_inc_q16` (per-sample phase step, 65536 = one cycle, tracks pitch bend), `amp_q15` (live envelope 0..32767) — so ADSR/bend/pan all work via AMY. State between calls = `static`s in the body; per-osc polyphonic state = static arrays indexed by `osc`.

Shipped example `tulip/fs/tulip/ex/c_dsp_demo.py`: bitcrusher + a CZ-101-style phase-distortion saw (`import ex.c_dsp_demo` or run it from the editor). Gotcha it documents: bind HIGH osc numbers (e.g. 200) — low oscs belong to midi.py's default synths and raw events to them are silenced.

Verified headless on the live render path: a muting user osc zeroes the bus during a held note; hot-swapping to the CZ body restores full-scale audio; effect path regression-tested.

bwhitman:
**It runs on the AMYboard.** 446042e2 + 18c1a984 land the ESP32-S3 slice:

- `tulip/shared/3rdparty/xcc700/` — vendored fork of [xcc700](https://github.com/valdanylchuk/xcc700) (MIT): adds `int16_t` pointers/arrays (`l16si`/`s16i`, ×2 indexing), casts-as-no-ops, `static` locals → function-scoped .bss (the user-DSP state idiom), and a callable `xcc700_compile()` API (in-memory ELF, error strings with line numbers, e_entry = the user function so the loader returns a ready pointer).
- `user_c_dsp.c` refactored to common install/dispatch + per-platform backends: libtcc (desktop) / xcc700t + `espressif/elf_loader` 1.3.1 (ESP, code loaded into **PSRAM** — zero IRAM pressure). On ESP the render hook chains CV first, then user oscs.

**Bench results** (local AMYboard, USB CDC REPL): the 10-bit crusher *body* compiled on-device to slot 0 and dispatched at exactly 172 blocks/sec while a synth note played (calls 345→517/s); the integer CZ phase-distortion osc — static per-osc phase array in PSRAM bss and all — bound to osc 200 and rendered a held note (calls 3624→4142). Same Python API, same effect bodies as desktop, no toolchain on the board.

Watch-item: one unexplained CDC drop during an install mid-session (no panic on the UART console; possibly host-side USB re-enumeration). Firmware size cost is fine — 26% app-partition headroom remains with elf_loader's default symbol tables in.

bwhitman:
adbde8f3: **API v2 per dpwe's review** — user code now runs directly on AMY's S8.23 buffers (zero-copy, no conversion pass, full bus headroom), and instead of hiding the fixed point, helpers are **exported into user code on both backends** (`tcc_add_symbol` on desktop, `esp_elf_register_symbol` on ESP):

- `cos_lut(phase_q23)` — AMY's cosine table (S8.23, phase normalized to one cycle)
- `fxmul(a, b)` — S8.23 multiply
- `to_int16(s)` / `from_int16(v)` — for people who think in 16-bit PCM

The shipped CZ example's inner loop is now exactly dpwe's suggestion: `buf[i] = fxmul(cos_lut(wp << 7), amp);` — AMY's real cosine instead of a parabola approximation. Osc contract: `render(int *buf, int frames, int osc, int phase_inc_q16, int amp)` with `amp` in S8.23. amy re-pinned to `5376e2f` (bus-dsp-hook merged with amy main to pick up `cos_lut`; the local `make test` deltas match pristine origin/main on this Mac — pre-existing platform float drift, CI will arbitrate).

**Bench-verified on the AMYboard**: crusher quantizes the live bus (verified numerically on desktop: every sample an exact multiple of 2^18 while enabled) and the cos_lut CZ osc renders a held note at 172 blocks/sec — i.e. runtime-compiled user code calling firmware-exported AMY functions through the ELF loader's symbol resolution, on device. That's the extensibility story working end to end: adding future exports (more tables, pitch helpers, RNG) is one line in `udsp_exports[]`.

bwhitman:
**Web slice, phases 1+2 of 3 done** (c6e98dc1 here, 0057f78 on amy#874):

**Phase 1 — `xcc700w`, a WebAssembly backend for the vendored compiler** (same small-C front end as the Xtensa one). The trick that makes it fast: the emitted module *imports* AMY's linear memory and gets `cos_lut` wired directly to the AMY module's export — user code reads/writes AMY's actual buffers and calls AMY's tables with **zero JS in the per-sample path**; `fxmul`/`to_int16`/`from_int16` are inlined as wasm intrinsics. Verified in Node: the shipped crusher body crushes shared-memory int32s bit-exactly, and the CZ phase-distortion body matches a JS reference model **sample-exactly** across multiple render calls, static per-osc bss state and all (crush.wasm = 575 bytes, cz.wasm = 1303 bytes).

**Phase 2 — AMY web hooks** (on the amy#874 branch): `amy_render_js_hook` / `amy_bus_postprocess_js_hook` EM_ASM variants following the existing `amy_block_processed_js_hook` pattern (the render one fills the old TODO at that site), plus web exports of `_cos_lut` and `wasmMemory`.

**Phase 3 — remaining plumbing** (next session): link xcc700w into the micropython web build behind `TULIP_USER_C_DSP`, bridge `install_c_process` bytes → main thread → AMY worklet (instantiate with `env.memory` = AMY's memory, bss/rodata from `_malloc`), a small JS slot table behind the two js hooks, then browser-verify via `webdev.py`.

bwhitman:
This totally works on hardware and web and desktop. There's more cleanup and docs but i will merge this now as it's a no-op unless you know about it.

PR #1153, opened 2026-07-13
Pin amy: fix sprint_event %hu abort under MicroPython internal printf (amy#888)

bwhitman:
Pins the `amy` submodule to [shorepine/amy#888](https://github.com/shorepine/amy/pull/888) and refreshes the vendored refdocs stamp.

## Bug

`sprint_event` (amy `src/patches.c`) printed the event `osc` field with `PRIu16` → `%hu`. MicroPython's `mp_vprintf` has no `h` length-modifier support, so wherever `MICROPY_USE_INTERNAL_PRINTF=1` (the default; desktop builds compile `shared/libc/printf.c` via `extmod.mk`) binds `snprintf` to the internal implementation, serializing any event hits `assert(*fmt == '%' || !"unsupported fmt char")` in `py/mpprint.c:568` and aborts (desktop builds run `DEBUG=1`, so the assert is live). Triggers: `tulip.amy_dump_state()`, `amyboard.update_sketch_knobs()`, anything else that serializes events.

## Who is actually affected

- **tulip/macos: NOT affected** — verified live at the desktop REPL (`tulip.amy_dump_state()` returns fine). Darwin's fortify headers rewrite `snprintf` calls to `___snprintf_chk` (system libc) even with the Makefile's `-U _FORTIFY_SOURCE` (Darwin's `_USE_FORTIFY_LEVEL` defaults to 2), so amy's calls never reach the MicroPython override (`nm patches.o` shows only `___snprintf_chk`).
- **tulip/linux: affected** — glibc fortify is genuinely disabled by `-U _FORTIFY_SOURCE`, so `snprintf` binds to the internal printf, and `DEBUG=1` keeps the assert live.
- **VCV Rack plugin build (in progress, not yet in-tree): affected** — worked around locally with `-DMICROPY_USE_INTERNAL_PRINTF=0`.

## Verification

- Harness linking the desktop build's own `mpprint.o`/`printf.o` and calling the internal `snprintf` directly: `"v%hu"` → `Assertion failed: (*fmt == '%' || !"unsupported fmt char"), function mp_vprintf, file mpprint.c, line 568` (abort, exit 134); `"v%u"` → `v3`, exit 0.
- amy `make test`: 109 tests pass; all amy#888 CI checks green.
- Tulip Desktop rebuilt at this pin: `tulip.amy_dump_state()` wirecode output byte-identical to before the fix.
- Audited amy + tulip/shared for other `h`-modifier formats reaching overridden functions: none (all remaining `PRIu16` uses are `fprintf(stderr, …)`, which is not overridden; amy redefines `PRIu8` to `"u"`).

## Note on the alternative fix (MICROPY_USE_INTERNAL_PRINTF=0 for desktop)

Not taken: upstream's unix port also keeps the internal printf (the `-U _FORTIFY_SOURCE` comment in our Makefiles is copied from it), and `tulip/shared` has real `printf()` callers (e.g. lodepng error reporting in `modtulip.c`) whose output currently routes through `mp_plat_print` to the Tulip console — disabling the override would silently divert those to raw stdout. Fixing the format string in amy is smaller and covers every port, including the VCV build without its local workaround.

## Before merge

- [ ] Merge amy#888, then re-pin this branch to the **merged** amy commit (and re-run `tulip/server/sync_amy_docs.py`) per the AMY submodule policy.


bwhitman:
Closing as part of the amy-pin cleanup consolidated into #1238.

Heads up, this one is **not** subsumed: it pinned amy to `cd5b35ea`, a commit on the branch for [amy#888](https://github.com/shorepine/amy/pull/888), which is **still open** and never landed on amy `main`. So the `sprint_event` `%hu` fix is *not* in the new pin. Per the AMY submodule policy tulipcc can't pin to an unmerged amy commit anyway, so this PR couldn't merge as-is. The fix itself is preserved in amy#888 — merge that in amy first, then bump the tulipcc pin.

PR #1164, opened 2026-07-16
Pin amy to d5c80a377 (amy#905)

bwhitman:
## ⛓️ Pin `amy` → `d5c80a377` (validates shorepine/amy#905)

`amy` was just merged to **shorepine/amy@main**. This PR bumps the `amy` submodule
pointer to that commit so the full tulipcc CI (AMYboard preview, Tulip preview, desktop
build) runs against it.

**Please test, then merge this PR manually** to move tulipcc onto the latest AMY.

| | |
|---|---|
| Merged AMY PR | shorepine/amy#905 — ESP32 RAM placement optimization from github.com/rt-rtos/S3-Amysynth/commit/9aba81a. |
| Author | @dpwe |
| Merged by | @dpwe |
| AMY commit | shorepine/amy@d5c80a3774ef172b7a9f712b9cc355fd3b6ec5eb |

### What changed in this AMY PR

- [`ee9fb15`](https://github.com/shorepine/amy/commit/ee9fb152d52694b61ac742021b42215a61ba2b0a) ESP32 RAM placement optimization from github.com/rt-rtos/S3-Amysynth/commit/9aba81a.

<details><summary>9 file(s) changed (+36/-22)</summary>

- `amy/headers.py` (+1/-1)
- `src/amy.c` (+1/-1)
- `src/amy.h` (+14/-0)
- `src/clipping_lookup_table.h` (+1/-1)
- `src/envelope.c` (+3/-3)
- `src/filters.c` (+5/-5)
- `src/log2_exp2.c` (+4/-4)
- `src/oscillators.c` (+6/-6)
- `src/patches.c` (+1/-1)
</details>

---
cc @dpwe — you merged the AMY change above; please review the tulipcc CI on this PR and merge it here when green.

PR #1165, opened 2026-07-16
Pin amy to 156fad9 (amy#907): #905 IRAM placement gated off for Tulip

bwhitman:
Tests the theory from the [PR #1164 HW CI failure](https://github.com/shorepine/tulipcc/actions/runs/29502489895): amy#905's IRAM/DRAM render-path placement hangs TULIP4_R11 at boot (console stops after `Starting MicroPython on core 1`, never reaches the REPL; AMYboard passed the same run).

[amy#907](https://github.com/shorepine/amy/pull/907) keeps #905's speedup for AMYboard but makes `AMY_IRAM_ATTR`/`AMY_DRAM_ATTR` no-ops when `TULIP` is defined. If HW CI passes both boards here, that confirms the IRAM move as the cause; then amy#907 can merge and the pin can move to its merge commit (superseding #1164).


bwhitman:
Superseded by #1163, which pins amy to latest main bfced59 (1.2.65 — includes amy#904, #905/#907 and the release bumps).

PR #1171, opened 2026-07-17
Adapt to amy#913: single-voice drum kits with one dedicated osc per drum

bwhitman:
Companion to shorepine/amy#913 (Dan's one-osc-per-drum restructuring of the drum kit patches). amy#913 is now merged and the `amy` submodule is pinned to its merge commit `2505e75` (tree-identical to the tested PR head `65f07d5`).

## What amy#913 changes

Drum kit patches (258, 384-390) are now **single-voice**: the one voice is a container holding a dedicated osc per drum sound (~38-42 oscs), each pre-loaded with its PCM preset. The note mappings address the per-drum osc (`v<k>n<base>l%vi%i`), osc params persist per drum across hits, and an osc-less synth command broadcasts to every osc in the voice. Any drum setup with `num_voices > 1` now over-allocates ~38 oscs per extra voice and must be fixed.

## tulipcc adaptations

**num_voices=1 everywhere a drum kit loads**
- `amyboard.py` boot/restart lines → `i10K384iv1` (device, VCV, web sim)
- Web editor: channel-10 drum default and the drum preset Load path
- `midi.py` default channel-10 `DrumSynth(num_voices=1)`; `DrumSynth` default arg
- World sketches (`acid_generator`, `house_generator`, `drum_machine`) — these also drop their `amp=5`, since the kit gain is baked into the mapping's velocity scale and an osc-less `amp` now broadcasts ×5 to every drum osc
- Docs (`docs/music.md`, `docs/amyboard/python.md`, `docs/amyboard/faq.md`) and `AMY_AGENTS.md` sketch-generation guidance

**Drum knob surface (`editor_drums.js`) reworked — no io rewrites at all** (final design per Dan's review)
Every per-drum knob is a persistent param on the drum's dedicated osc, one note-addressed line per mapped note — `i10n39f523.251a0.425Q0.1G0F16000R0.7P0.0000956` — routed through amy#913's velocity-less "fake note on" so the UI never needs osc numbers. `f` = pitch (freq const, 440 Hz = no shift), `a` = vol × channel level, `Q` pan, `G/F/R` filter, `P` offset (persistent trigger_phase). The kit's `midi_note_cmd` strings are never rewritten; the kit patch string is parsed only to group the kit into named UI rows.

**Drum channel Level drops `v0`**
A kit has no control osc anymore, so the Level slider / CC 7 sends the osc-less broadcast `i<ch>a<val>` (melodic channels keep `i<ch>v0a<val>`). Since each drum osc's amp is vol×level, a level change then re-sends every sound's per-note line with the new product (debounced); in the knob log the broadcast replays before the per-note lines, which also lets the loader divide recorded amps back into vol knob positions. Known tradeoff (ack'd by Dan): a hardware CC 7 sweep plays only the broadcast, transiently flattening per-drum vol ratios — workaround is drums on their own bus with the FX level slider.

**Legacy sketch migration**
- `i<ch>K<kit>iv6` setup lines are normalized to `iv1` at load (web `rebuild_knob_log_from_block` + `amyboard.py _apply_knobs_text`, covering device flash sketches)
- Pre-#913 io/level lines (and this PR's interim io-rewrite form) are converted to the per-note form when the drum surface renders (values recovered, kit re-sent to restore pristine mappings, stale slots dropped); `amyboard.py` skips replaying stale io lines, whose osc-less templates would broadcast a hit to every osc in the voice
- Knob-log structural keys keep a leading `i<ch>n<note>` literal so each drum's osc-param line gets its own last-wins slot

**Regenerated:** `patches.generated.js` (new kit strings) and the refdocs vendored-from marker.

## Tested (web sim, amy @ 65f07d5 == merged 2505e75)

- Ch 10 boots the TR-808 kit as `if3iv1in38Z` + per-drum oscs; hits play, no console/AMY errors
- Clap knobs (pitch +3 st, vol 0.8, pan 0.1, offset 0.25) produce one line `i10n39f523.251a0.425Q0.1G0F16000R0.7P0.0000956`; AMY readback shows it all landed on the clap's osc (`v4w7p3a0.425f523.249...Q0.100P...`) and values round-trip into the knobs exactly
- Level move to 0.531: untouched kick lands at `a0.531`, vol-0.8 clap at `a0.425`; broadcast line replays first; no io lines anywhere in the log
- Legacy block (`iv6` + old io + `v0a` level) loads normalized and migrates with knob values recovered (pan 0.9, cutoff 2000, res 0.7)
- Drum preset Load (TR-909) emits `K385i10iv1`
- HW CI (physical bench): PASS on both AMYboard and Tulip

Follow-up filed: shorepine/amy#916 (PCM pitch change while a note is sounding glitches instead of bending).


dpwe:
Note: PCM `trigger_phase` (`P`) is no longer consumed by note-ons, it's a persistent property of the osc, and should be set on the osc, NOT as part of the note-on command.

Pitch is less obvious; I think it's more natural to change the `freq_coef[0]` away from 440 Hz (rather than changing the per-drum base_note value in the `io` commands.

If we could also migrate the per-note amplitude scaling to elsewhere in the static coefficients (like, `amp_coef[VEL]`, with suitable pre-compensation for the 60db 0-to-1 scale, baked in to the patch via `headers.py`), then the `midi_note_cmd` strings could be left untouched for any param changes (and, I think, the web editor code would not need to parse the patch string, which seems fragile).

bwhitman:
[claude]

@dpwe Addressed in c3c8f71 (amy re-pinned to your `65f07d5`):

- **offset** now sets the osc's persistent `trigger_phase` in the per-note osc-params line — no `P` in the io template anymore.
- **pitch** now sets the osc's freq const (`f440*2^(st/12)`) instead of touching the mapping's base note; recovered on load as `12*log2(f/440)`.

So a drum edit is now one persistent osc-params line per note, `i10n39f523.251Q0.1G4F2000R0.7P0.0000956`, addressed by note through the fake-note-on path — verified in the web sim that the values land on the drum's osc (`v4w7p3f523.249...Q0.100P...`), survive hits, and round-trip back into the knobs.

The **only** remaining io rewrite is vol, since the kit-balancing gain lives in the mapping's velocity max — the template itself is kept byte-identical to the kit's canonical `v<k>n<base>l%vi%i` form. If you move the per-note gain into the osc's amp coefficients in headers.py (your `amp_coef[VEL]` pre-compensation idea), I'll happily delete that last rewrite; the one wrinkle to watch is that the channel Level is now the osc-less broadcast `i10a<val>`, which writes `amp_coef[CONST]` on every drum osc — so per-drum vol needs to live in a coefficient the Level broadcast doesn't overwrite (VEL with baked pre-comp works, CONST doesn't).

On patch-string parsing: agreed it's fragile in principle, though the editor would still need *some* kit metadata (note→preset/osc grouping for the row labels, per-sound audition, and the vol base gain). If amy ever exposes that mapping programmatically (or via the readback API), I can drop the string parse.

dpwe:
I'm happy to report that changing the `freq=` away from the default of 440 works fine for changing the pitch of PCM voices.  So I recommend this for implementing the pitch knobs on the AMYboard web drum editor. 

I'm still looking at storing the per-drum amplitude scaling into the `amp_coef[VEL]`.

dpwe:
I would like this change to avoid rewriting the midi_note_cmds (`io`) altogether. Instead, all the per-drum-sound knobs modify the params of the underlying oscillator: "offset" is directly `trigger_phase`, "pitch" sets `freq` above or below its standard 440Hz value, and "vol" is combined with the overall channel level to set the osc's `amp`. 

Note that since each drum osc will now potentially have a different `amp`, changing the overall channel level requires rewriting each of those `amp`s individually as the product of the channel level and that drum's "vol" knob. 

I think this change makes it unnecessary for the web editor code to parse the drum patch setup at all (it no longer needs to extract the initial `io` parameters) but I'm not sure.

bwhitman:
@dpwe Done in 6dea4db — **no io rewrites anywhere anymore**. Every per-drum knob is a persistent param on the drum's osc, one note-addressed line per mapped note:

```
i10n39f523.251a0.425Q0.1G0F16000R0.7P0.0000956
```

(`f` = pitch, `a` = vol × channel level, `Q` pan, `G/F/R` filter, `P` offset.)

Level works as you described: the slider/CC 7 still sends the osc-less broadcast `i10a<level>` for live response, then (debounced) re-sends every sound's per-note line with the new `vol*level` product — verified in the sim that an untouched kick lands at `a0.531` and a vol-0.8 clap at `a0.425` after a level move to 0.531. In the sketch log the broadcast line always replays before the per-note lines, which also lets the loader divide the recorded products back into the vol knobs. One caveat: a *hardware* CC 7 sweep only plays the broadcast (a device ic template can't fan out 38 different values), so it transiently flattens vol ratios until the next per-note send or sketch replay — web-driven level moves are always exact.

On patch parsing: you're right that no knob value depends on it now. The editor still parses the kit string once, but only to *group* the kit into named rows (note → osc → preset, joined with the sample-name/frames table) and pick audition notes. If amy ever exposes the note→preset map programmatically I'll drop that too.

Migration handles all three older recorded forms (pre-#913 baked templates, this PR's earlier velocity-max vol lines, and the `v0a` level).

dpwe:
Ok, sounds good. Point taken about CC7 being unable to preserve unequal gains across drums, I guess we'll live with that for now. A work-around for users is to put the drums on their own bus and control their overall level with the FX level slider.

PR #1174, opened 2026-07-17
Drum knobs install real device MIDI-CC (ic) mappings; pitch knob uses native AMY values

bwhitman:
Follow-up to #1171. Two fixes to the drum knob surface's MIDI CC story:

## Device-side ic mappings for drum knobs

Assigning a MIDI CC to a drum knob (clicking the knob title) now installs a real device-side `ic` mapping, exactly like the melodic knobs — previously drum knobs were `no_device_cc` (the old io-template design couldn't nest inside an ic template) and CC only worked through the web's `move_knob` routing. Bass-drum pitch on CC 77 installs:

```
i10ic77,1,220,880,0,i%in35f%vZ
```

- Each ic template carries **only that knob's parameter**, note-addressed via the fake-note-on path.
- Multi-note sounds (the toms: 41/43/45/66) get a Z-joined multi-command template (`i%in41f%vZi%in43f%vZ...`), the same pattern K257 uses for its dual-osc knobs.
- The cutoff template pins `G4` (LPF24) so a CC sweep is audible from any state.
- The vol template plays the osc amp at the raw CC value — an ic mapping can't apply the web knob's vol×level product.
- The offset knob's mapping converts its 0..0.95 UI fraction into the sample's absolute trigger_phase range (new `cc_min_value`/`cc_max_value` overrides consumed by `build_knob_cc_value`).
- Knob-log structural keys learn the `i%in<note>` form so pitch-CC lines for different drums get their own last-wins slots (they'd otherwise all normalize identically and collapse onto one line).
- The recorded `i<ch>ic...` lines ride the sketch's knob block, so device boots replay drum CC maps like any other mapping.

## Pitch knob → native AMY values

The pitch knob's -24..+24 semitone range is replaced with the AMY-native freq const: **log 220..880 Hz, 440 = no shift** (±1 octave). The knob's value is exactly what goes on the wire (`f%v`), which is also what makes the log-scaled ic mapping above work with no unit conversion. Recorded `f` values load back directly; pre-amy#913 semitone pitches still migrate via `440*2^(st/12)`.

## Tested (web sim)

- CC-assign on bass-drum pitch records + sends `i10ic77,1,220,880,0,i%in35f%v`
- Injected hardware CC (`0xB9 77 127`) drives the kick's dedicated osc to `f880.000` via AMY's own mapping
- Pitch knob defaults to 440, a move to 660 records `i10n39f660a1Q0.5G0F16000R0.7P0` and round-trips back into the knob
- Tom pitch template Z-joins all four mapped notes; offset CC max computes to the sample's phase range
- No console/AMY errors; melodic-knob CC paths untouched (K257 ic lines key identically to before)


bwhitman:
Also fixed in e84a280 (rode along here since it's the same surface): the startup errors
```
cannot find 38 oscs for patch 384 for voice 10. not setting this voice
```
came from two drum boot sends the #1171 `iv6`→`iv1` sweep missed — they build the wire string by concatenation (`"i" + DRUM_CHANNEL + "K" + DRUM_DEFAULT_PATCH + "iv6Z"`), so the literal-string grep didn't catch them: `start_audio`'s pre-`run_sketch` init and the first-run (no sketch yet) default in spss.js. Each asked for 6 voices × ~38 oscs; drum voices 10–11 didn't fit the osc pool. amyboard.py's own `i10K384iv1` rebuilt the synth right afterward, which is why drums still played. Both now send `iv1` — verified a clean boot with no allocation errors and the kit at `if3iv1in38Z`.

bwhitman:
Added in 8bff24e (with its AMY half in shorepine/amy#918): **all Level sliders now use a per-instrument level, `i<ch>iV<val>`**, instead of sending amp to osc 0 of the synth.

AMY side (amy#918): `synth_level` / wire `iV` — float, default 1, stored on the instrument, applied at render to every osc of the synth. Applied at the **audio output stage** (pre pan/bus mix), not per-osc amp — scaling amps squares the level on layered patches whose control oscs multiply into their carriers (measured, then fixed). O(1) per-osc lookup via a per-voice level mirror. All 114 amy tests pass (incl. a new `TestSynthLevel`; level ratios exact at 0.4999×/2.0002× for iV 0.5/2); the K257 patch's baked CC 7 mapping becomes `i%iiV%v` with bit-identical test output.

tulipcc side:
- Slider + CC 7 knob send `iV` on **every** channel (melodic + drums); the web slider's readback comes from the synth preamble (`if3iV0.531iv1in38Z`)
- Drum vol knobs are now bare osc amps — level is orthogonal, so the debounced vol×level product rewrite from the earlier commit is deleted (~50 lines gone)
- Migration: pre-iV level lines (`v0a` melodic, the interim `i<ch>a` broadcast) move to `iV`, and interim per-note amp products divide back into bare vols — verified a synthetic interim-era block converts to `i10iV0.5` + `a0.8` with knobs recovered

⚠️ Merge ordering: this PR now depends on amy#918 — same dance as #1171 (merge amy#918 first, re-pin here to its merge commit, then merge this).

PR #1192, opened 2026-07-18
amyboard_fs_create: flash the Gamma9001 drums partition

mateusz28011:
## Summary

`tulip/amyboard/amyboard_fs_create.py` never packs `drums.bin` into the `drums`
partition, even though the partition table declares it. This ports the same
conditional block `tulip/fs_create.py` already uses (added in #1084).

## Why

`fs_create.py amyboard` (what release CI runs) packs drums, so released firmware is
fine. But `docs/amyboard/firmware.md` and `docs/amyboard/faq.md` both point people at
`amyboard_fs_create.py` for local rebuilds, and that path leaves the `drums` partition
erased. The missing-partition stderr note from #1084 doesn't fire here — the partition
*entry* exists, so AMY mmaps the erased region and Gamma9001 presets (256–391) just
play nothing.

Looks like an oversight from #1084: its description covers `fs_create.py` only.

## Verified

- `python3 -m amy.headers gamma9001` produces `drums.bin` (3,735,788 bytes), within
  the partition's 3,801,088-byte capacity.
- A full image built with this block carries `drums.bin` byte-identical at `0xc60000`.

### DIFF
--- a/tulip/amyboard/amyboard_fs_create.py
+++ b/tulip/amyboard/amyboard_fs_create.py
@@ -83,10 +83,31 @@ with open("build/amyboard-sys.bin","wb") as fh:
 print("... done.")
 
 
+# Gamma9001 drum banks: generate drums.bin from the amy submodule and flash it
+# into the `drums` partition (AMY mmaps it at boot; see amy_connector.c).
+drums_partition = None
+try:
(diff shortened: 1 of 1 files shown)


bwhitman:
Nice catch thank you

PR #1200, opened 2026-07-19
Restore tulip.amy_dump_state removed with the zA path (broke amy.dump_state() off CPython)

mateusz28011:
## Summary
#1190 removed the zA control command and, with it, `tulip.amy_dump_state`
(`modtulip.c`) and the amyboardweb WASM bridge (`spss.js`), on the rationale
that shorepine/amy#934 exposes this as `amy.dump_state()`. But `amy.dump_state()` is not
self-contained off CPython — it looks up a module-global `_dump_state`
(`return _dump_state()`), and on non-CPython builds that global is meant to come
from `tulip.amy_dump_state`. Removing the tulip binding leaves it undefined, so
`amy.dump_state()` raises `NameError: name '_dump_state' isn't defined`. CPython
(`c_amy.dump_state`) is unaffected.

The exact failure differs by platform, and both matter:

- **Native MicroPython / ESP32:** the `tulip.amy_*` functions are real native
  bindings, so `amy/__init__.py`'s import-time try block binds the first four
  and then fails specifically at `_dump_state = tulip.amy_dump_state` — the one
  entry #1190 deleted.
- **Web / WASM:** every `tulip.amy_*` native is `#ifndef __EMSCRIPTEN__`-excluded,
  so that try block has always failed at its first line and none of these are
  bound at import; they're patched in from JS at runtime after import instead.
  #1190 removed the runtime `tulip.amy_dump_state` JS patch. And `amy._dump_state`
  was never patched directly on web, so restoring only `tulip.amy_dump_state` is
  not enough there — `amy.dump_state()` still NameErrors until amy's own global
  is set (see the spss.js note below).

## Changes
- **`modtulip.c`** — restore `tulip_amy_dump_state` (function body byte-identical
  to the pre-#1190 version, plus one explanatory comment) and its module-table
  entry, inside the existing `#ifndef __EMSCRIPTEN__` guard.
- **`spss.js`** — restore the `amy_dump_state_to_string` cwrap, the
  `amy_dump_state_js` bridge and its `registerJsModule`, and the
  `tulip.amy_dump_state = _amy_dump_state_js` patch. Also set
  `amy._dump_state = _amy_dump_state_js`: `amy`'s import-time bind can't succeed
  on web (above), and `dump_state()` resolves `_dump_state` as a module global at
  call time, so patching amy's own global is what makes `amy.dump_state()` work.
  This is analogous to the adjacent `amy.ticks_ms = amy_sysclock` patch, done for
  the same import-ordering reason — though note that one overrides the public
  function while this overrides the private `_dump_state` indirection.

Pure restore of the dump-state readback path; the dead zA authoring chain #1190
removed stays gone.

## Testing
- **AMYboard Web (Simulate), clean boot, no manual patching:** both
  `amy.dump_state()` and `tulip.amy_dump_state()` return the state dump
  (116-char FX dump for bus 0); reproduced the `NameError` before the fix.
- **amy `make test`:** 117/117 pass (incl. `TestDumpState`).
- **CPython:** `amy.dump_state()` returns a dump via `c_amy`.
- **macOS Desktop:** builds and links; `tulip_amy_dump_state` /
  `amy_dump_state_to_string` present in the binary.
- Not run locally: ESP32 firmware build, hardware smoke test.

### DIFF
--- a/tulip/amyboardweb/static/spss.js
+++ b/tulip/amyboardweb/static/spss.js
@@ -18,6 +18,7 @@ var _url_env_pending = false;
 var _url_env_autoplay = false;
 var _url_env_loaded = false;
 var amy_yield_synth_commands = null;
+var amy_dump_state_to_string_c = null;
 
 var _python_error_buffer = "";
 var _python_error_timer = null;
@@ -949,6 +950,9 @@ if (typeof amyModule === 'function') _amy_wasm_ready = amyModule().then(async fu
   );
   amy_yield_synth_commands = am.cwrap(
     'yield_synth_commands', 'number', ['number', 'number', 'number', 'boolean', 'number']
+  );
+  amy_dump_state_to_string_c = am.cwrap(
+    'amy_dump_state_to_string', 'number', ['number']
   );
   amy_start_web_no_synths();
   amy_module = am;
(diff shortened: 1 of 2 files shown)


bwhitman:
Yes thanks for this! We do want to this to get fixed but are discussing a more holistic approach so we stop running into this whenever we update the AMY C API . Be back soon with a potential solution

bwhitman:
We built this in #1201 for all AMY C API calls, for our 4 non-C platforms, so we avoid having to do this manually. In about 20 minutes you can do `amy.dump_state()` from any Tulip or AMYboard surface and it should just work . Thank you for the report!

PR #1208, opened 2026-07-20
amyboardweb: regenerate amy_api.generated.js for pinned amy aa48a5a

mateusz28011:
The committed `tulip/amyboardweb/static/amy_api.generated.js` no longer matches the
pinned amy: it still carries `voices` (wire `"r"`) in both `AMY_KW_MAP` and
`AMY_KW_PRIORITY`, which shorepine/amy#943 removed. Everything after it in the
priority table is shifted by one as a result.

Regenerated with amy's own generator against the current pin:

    python3 amy/scripts/gen_amy_js_api.py    # → build/amy_api.generated.js

That output is byte-identical to what `tulip/amyboardweb/dev.py` copies into
`static/` during a build, so this only commits what any local build already
produces. `patches.generated.js` was checked the same way and is still in sync.

Deploys were never affected — the release build regenerates and copies the file
before staging. What it does affect is any local build: it silently dirties the
tree, which is how this was found.

Follow-up filed separately: nothing currently detects this drift, because the
amy-pin bot bumps the gitlink without a checkout (by design) and so can't
regenerate derived files.


bwhitman:
Another good instinct, but instead of fixing this as a one off we'll fix this "for real" as a deploy so these generated files don't fall out of sync. Track it in https://github.com/shorepine/amy/pull/953

PR #1249, opened 2026-07-27
ci: persist the esp-idf ccache across runs

bwhitman:
Follow-up to #1248. That one killed the 340s submodule clone; this goes after the other half — the ~420s esp-idf compile.

`IDF_CCACHE_ENABLE=1` was already set on the preview build (that's *why* the two firmware targets share one job — the second hits the warm cache). But it started cold every run, because nothing carried the cache between runs.

## Two traps this had to avoid

**1. `~/.ccache` is the wrong path, and would have failed silently.** The compile runs as root *inside* the esp-idf container, where `$HOME` is the container's own `/root`. An `actions/cache` on the runner's `~/.ccache` would have archived an empty directory, reported success, and shown a 0% hit rate forever with nothing obviously wrong.

`CCACHE_DIR` now points into the bind-mounted workspace, computed as `$(cd ../.. && pwd)/.ccache` from the build directory — relative, so it doesn't depend on where `esp-idf-ci-action` chooses to mount things (currently `/app/shorepine/tulipcc`, which is not what you'd guess).

**2. Cache scope.** A cache saved from a PR branch is visible only to that branch — not to other PRs, not to main. PR runs saving their own would be pure upload time for nobody. But PR runs *can* read the default branch's caches. So:

- **PR previews restore only** (`actions/cache/restore`)
- **The two release firmware jobs restore and save** — they run on `main`, so they're the only ones that can populate a cache PRs will actually see

Both release jobs share one key namespace deliberately: each restores the newest entry and saves a superset, so the single cache accumulates objects for both boards over successive merges. Concurrent runs can race and one save drops the other's additions — self-heals on the next run.

## Also

- `.ccache` is pre-created as the runner, so a cache miss doesn't leave ccache to create it root-owned `0700` and break the save.
- Chowned back before saving, since the container wrote as root.
- `ccache -s` at the end of each build so the hit rate is visible in the log. Wrapped in `|| true` — a stats command should never be able to fail a build.
- `CCACHE_MAXSIZE=800M`, comfortably above the ~200-300MB two board configs should produce, while keeping the cache upload/download quick.

## What to expect from *this* PR's run

**Roughly nothing, and that's the expected result.** No cache exists under the new key yet, and this PR can't create one PRs can read — only a merge to `main` can. So the preview here should be a normal-speed cold build; what it proves is that the plumbing doesn't break anything and that `ccache -s` reports sane numbers.

The payoff shows up on the *next* PR after this merges and one release run has populated the cache. Most of the ~1400 objects in a firmware build are IDF components (mbedtls, lwip, freertos, esp_wifi) that don't change between PRs, so the hit rate should be high for anything not touching `tulip/shared`.

I'd suggest merging on the strength of a green run plus the `ccache -s` output, then checking the *following* PR's build time to see the actual win.

### DIFF
--- a/.github/workflows/amyboard-pr-preview.yml
+++ b/.github/workflows/amyboard-pr-preview.yml
@@ -148,10 +148,24 @@ jobs:
           restore-keys: esp-idf-ccache-v5.4.1-
 
       # --- Firmware (.bin sets) ---
-      # Build AMYBOARD then TULIP4_R11 in ONE esp-idf container: the two targets
-      # share most sources, so the second build hits ccache for the common units
-      # (we enable IDF_CCACHE_ENABLE) instead of recompiling from scratch. The
-      # Tulip firmware is what the HW CI bench flashes onto the physical Tulip
+      # Build AMYBOARD then TULIP4_R11 in ONE esp-idf container.
+      #
+      # This comment used to claim the second target hits ccache for the units
+      # the two boards share. Measured on the cold run of #1249, that is false:
+      # 14 hits out of 2829 cacheable calls (0.49%). The boards share sources but
+      # not compile *flags* -- different -D sets (AMYBOARD/GAMMA9001/MAKERFABS vs
+      # the TULIP4_R11 set), different build/config include paths, different
+      # -fmacro-prefix-map -- and any of those changes the hash. So there is
+      # effectively no cross-board reuse, and the real ccache win is cross-RUN
+      # (same board, same flags), which is what the cache steps above are for.
+      #
+      # The single job still earns its keep for other reasons: both firmwares
+      # have to land in one workspace to be bundled into stage/ and deployed, and
+      # splitting would pay checkout + submodule init twice. Worth revisiting as
+      # parallel jobs if firmware time dominates again -- the ccache rationale
+      # for keeping them together no longer holds.
+      #
+      # The Tulip firmware is what the HW CI bench flashes onto the physical Tulip
       # (it shares the AMYboard bench); see tulip/amyboard/hwci/tulip_hwci.py.
       - name: Build AMYboard + Tulip firmware
         uses: espressif/esp-idf-ci-action@v1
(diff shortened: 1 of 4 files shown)


bwhitman:
## Cold-run result — plumbing works, plus one finding worth flagging

```
Cacheable calls:   2829 / 2829 (100.0%)
  Hits:              14 / 2829 ( 0.49%)
  Misses:          2815 / 2829 (99.51%)
Cache size (GB):  0.1 /  0.8 (11.67%)
```

Restore reported `Cache not found for input keys` — expected, nothing exists under the new key until this merges and a release run saves one. `-- ccache will be used for faster recompilation` confirms `CCACHE_DIR` landed somewhere ccache actually used, which was the main thing this run had to prove. 0.1 GB for both boards means `CCACHE_MAXSIZE=800M` is comfortable and the cache upload will be quick.

### The finding: cross-board ccache sharing was a myth

The workflow comment has claimed that building both boards in one container lets the second hit ccache for the shared units. **14 hits out of 2829 says otherwise.** The two boards share sources but not compile *flags* — different `-D` sets (`AMYBOARD`/`GAMMA9001`/`MAKERFABS` vs the TULIP4_R11 set), different `build/config` include paths, different `-fmacro-prefix-map` — and any one of those changes the hash. So there was effectively never any cross-board reuse to lose.

Corrected the comment in a09c2c78 rather than leaving a load-bearing claim that measurement contradicts.

Two consequences:

1. **This PR should help more than the in-run cache ever did**, not less. Cross-*run* reuse is same-board, same-flags, same paths — the case that actually hashes identically.
2. **Splitting the boards into parallel jobs is more attractive than I said earlier.** I dismissed that idea specifically because it would break ccache sharing between the two targets. There is no sharing to break. It's still not free — both firmwares have to reach one workspace for bundling, and you'd pay checkout twice — but the objection I raised was wrong, and it's worth revisiting if firmware time dominates after this lands.

### Verifying the actual win

This PR can't demonstrate it: PR runs only read caches from the default branch, so the payoff appears on the *first PR after this merges and one release run has populated a cache*. Suggest merging on green + these stats, then watching the next PR's `ccache -s`.

PR #1250, opened 2026-07-27
TEST ONLY — do not merge: measure the ccache from #1249

bwhitman:
**Throwaway PR. Not for merge** — it exists to measure the cross-run ccache from #1249 against a real C change.

Changes exactly one line in one leaf `.c` file, so the build has a single genuine cache miss to work with. The change is real but inert: `tsequencer_init()` cleared `defer_callbacks` and `defer_sysclock` but not `defer_args`, while the cleanup path in `tulip_amy_sequencer_hook()` clears all three. Harmless today (the array is zero-initialized, and `defer_args[i]` is only read when `defer_callbacks[i]` is non-NULL). Picked because it's one line in one translation unit and defensible rather than noise.

## This run is an unintended A/B, because I shipped a bug in #1249

Both release firmware jobs key their cache on `esp-idf-ccache-v5.4.1-${{ github.sha }}`. On a push that triggers both, that is the **same SHA**, so the second save always collides with the first and is dropped. Not the occasional race I described in #1249 — a guaranteed one, every push. From the release runs on `693d29cd`:

- **AMYboard release** — `Cache saved with key: esp-idf-ccache-v5.4.1-693d29cd…` (33.48 MiB, 1267 objects)
- **Tulip firmware release** — finished 37s later, **no** "Cache saved" line; its 1562 objects were discarded

So the only cache in the repo holds **AMYBOARD objects only**. Which means this one preview run builds:

- **AMYBOARD** — against a warm cache
- **TULIP4_R11** — against nothing

Same runner, same commit, back to back. That's a cleaner isolation of the ccache effect than a before/after across two runs would have given, so I'm taking the measurement before fixing the key.

Fix (per-board key namespaces, and the preview restoring both) goes up separately.

### DIFF
--- a/tulip/shared/tsequencer.c
+++ b/tulip/shared/tsequencer.c
@@ -39,7 +39,7 @@ void tulip_amy_sequencer_hook(uint32_t tick_count) {
 
 void tsequencer_init() {
     for(uint8_t i=0;i<SEQUENCER_SLOTS;i++) { sequencer_callbacks[i] = NULL; sequencer_period[i] = 0; sequencer_tick[i] = 0; }
-    for(uint8_t i=0;i<DEFER_SLOTS;i++) { defer_callbacks[i] = NULL; defer_sysclock[i] = 0; }
+    for(uint8_t i=0;i<DEFER_SLOTS;i++) { defer_callbacks[i] = NULL; defer_sysclock[i] = 0; defer_args[i] = NULL; }
     #ifndef AMY_IS_EXTERNAL
     amy_global.config.amy_external_sequencer_hook = tulip_amy_sequencer_hook;
     #endif


bwhitman:
## Result: 2.3× on the cached board

Cache restored: 33 MB, `esp-idf-ccache-v5.4.1-693d29cd…` — AMYBOARD objects only, per the collision described above. So this run is a within-run A/B: same runner, same commit, one cached board and one uncached.

### Compile phase, per board

| Board | Cold (#1249 run, no cache) | This run | |
|---|---|---|---|
| **AMYBOARD** (cache present) | 119s | **52s** | **2.3× faster** |
| TULIP4_R11 (no cache) | 140s | 184s | — |
| whole build step | 344s | 300s | |

Job total 8m16s, against ~10m for the cold runs.

TULIP4_R11 going *up* 44s is the honest caveat: single sample, so runner variance, plus a cold ccache pays a small write cost per miss. I would not read it as a regression, but I also can't rule one out from n=1.

### Reading the hit rate correctly

```
Cacheable calls:   4096 / 4096 (100.0%)
  Hits:            1278 / 4096 (31.20%)
```

**That 31.2% understates it, because ccache's counters live in the cache directory and were restored along with it.** The 4096 includes the 1267 calls from the release run that created the cache (1267 + 2829 = 4096 exactly).

This run alone: **1278 hits / 2829 calls = 45.2%** — which is essentially "the entire AMYBOARD build hit, the entire TULIP4_R11 build missed", exactly what one board's cache predicts.

### Extrapolating (estimate, not measured)

TULIP4_R11 has 1562 objects to AMYBOARD's 1267. Scaling the observed 52s by object count puts a warm TULIP4_R11 phase around 65s, which would take the build step from 344s cold to roughly 180s once #1251 lands and both boards have entries. Flagging that as arithmetic on one data point, not a measurement.

### Not for merge

Per the original ask. The one-line change in it is real and harmless if you ever want it, but nothing here needs to land.

PR #1256, opened 2026-07-28
amy.message() in C: ~15x faster wire-string building on every target

bwhitman:
## What

`amy.message()` costs hundreds of microseconds per message on device — almost none of it the actual work: it slices the keyword map twice per argument, calls a handler per argument, and accumulates with `+=`, and every one of those allocates on the MicroPython heap. This PR adds `tulip.amy_message()` (in `amy_connector.c`), the same function as a C table walk into a stack buffer, and installs it over `amy.message` from `_boot.py` on **every target** — TULIP4_R11, AMYBOARD, Tulip Web, AMYboard Web, desktop, and the VCV build — so every `amy.send()`, sequencer event, drum-machine knob, and note-on gets it.

Measured on desktop: **1.5 µs vs 24 µs** a message. The win is larger on ESP32-class hardware where heap allocation is the expensive part of the Python path.

## Correctness

- **Every formatter calls MicroPython's own, not libc's.** `mp_format_float()` is literally what `'%.6f' %` runs and `mp_obj_print_helper(PRINT_STR)` is literally what `str()` runs, so matching the Python builder byte-for-byte is structural, not sampled-for. (A `snprintf("%.6f")` version disagrees as soon as a value needs more digits than a float mantissa carries: `123.456789` → snprintf `123.456795`, MicroPython `123.456792`.)
- All five argument types are covered — int, float, str, list, ctrl-coefs — including amy's own asymmetries (a scalar coef truncs while list coef elements `str()`; lists join tuples but coefs assert on them; `'%'`-token passthrough), reproduced rather than tidied.
- **The C side returns `None` for anything it does not reproduce exactly** (unknown keyword, a numeric string where an int was meant, an overlong message) and the wrapper falls back to the original Python `amy.message`, which stays the definition of the wire format — and raises whatever it was going to raise. Kwarg combos that `message()` warns about under `show_warnings` go straight to Python so those diagnostics still fire; `insert_time` is applied in the wrapper exactly as the Python applies it.
- Verified against the Python builder on desktop: a ~50-case corpus + 25-value float sweep, **0 mismatches**, including bail cases and the fallback-raise path.

## The keyword table is generated, and CI-guarded

`tulip/shared/amy_kwmap.h` is generated from amy's `_KW_MAP_LIST` — the same source of truth the JS and GDScript maps are generated from — by `tulip/shared/gen_amy_kwmap.py`, which sources the **pinned** amy commit (local submodule when it sits at the pin, else GitHub raw), same pattern as `sync_amy_docs.py`. It's committed so none of the build systems needs its own codegen step, and the `amy-pin-check` workflow now runs `gen_amy_kwmap.py --check` on any pin move so it can't go stale. Table keywords are matched as qstrs interned once via `qstr_find_strn()` — referencing `MP_QSTR_` constants from a generated table collides with the frozen modules' qstr pool at build time.

## Builds

- macOS desktop ✅ (plus the corpus test above against the built app)
- `idf.py -DMICROPY_BOARD=AMYBOARD build` ✅
- `idf.py -DMICROPY_BOARD=TULIP4_R11 build` ✅
- Tulip Web (`tulip/web/build.sh`) ✅
- AMYboard Web (`tulip/amyboardweb` make) ✅

On web the builder compiles in unguarded (it's pure string building, no AMY link), and the JS side only overrides `amy.override_send`, so the install composes cleanly with the external-AMY bridge.


dpwe:
> including amy's own asymmetries (a scalar coef truncs while list coef elements str(); lists join tuples but coefs assert on them; '%'-token passthrough), reproduced rather than tidied.

I say tidy. Except '%'-token passthough, of course.  So trunc all floats, and accept tuples in place of lists.

dpwe (approved):

bwhitman:
@dpwe Tidied, in both places:

- **amy side** — shorepine/amy#1002: coefs now accept tuples like lists do, and every float element truncs to the same canonical form a scalar coef gets (one shared `elem_to_str()`); `'%'`-token passthrough untouched. Wire strings only get shorter — visible in the regenerated `patches.h` (`F200,1,,,5` where it used to say `F200.0,1.0,,,5.0`). 119 amy tests pass.
- **C side** (6fe4ccb) mirrors it: one `amy_put_elem()` shared by the list and coef paths, tuples accepted for both.

The amy pin here moves to the PR head temporarily so the C and the Python reference stay in lockstep; it needs repinning to the merged amy commit once #1002 lands, before this merges. Desktop corpus re-run against the new pin: 0 mismatches, now including tuple coefs and canonical float elements.

PR #1260, opened 2026-07-29
Update music.md for new PCM looping syntax.

dpwe:
Reflects changes in AMY PR [#1006](https://github.com/shorepine/amy/pull/1006)

### DIFF
--- a/AMY_AGENTS.md
+++ b/AMY_AGENTS.md
@@ -31,7 +31,7 @@ transfer_file disk_sample algorithm chorus reverb echo patch
 external_channel portamento sequence tempo synth pedal synth_flags num_voices
 oscs_per_voice to_synth grab_midi_notes note_source synth_delay preset
 num_partials start_sample stop_sample bus midi_cc midi_note_cmd cv_trigger
-patch_string
+patch_string wave_submode
 ```
 
 The keys *inside* a coefficient dict (e.g. `duty={'const':…, 'mod':…}`) are a **separate**
--- a/AMY_AGENTS.md
+++ b/AMY_AGENTS.md
@@ -31,7 +31,7 @@ transfer_file disk_sample algorithm chorus reverb echo patch
 external_channel portamento sequence tempo synth pedal synth_flags num_voices
 oscs_per_voice to_synth grab_midi_notes note_source synth_delay preset
 num_partials start_sample stop_sample bus midi_cc midi_note_cmd cv_trigger
-patch_string wave_submode
+patch_string mode
 ```
 
 The keys *inside* a coefficient dict (e.g. `duty={'const':…, 'mod':…}`) are a **separate**
--- a/docs/music.md
+++ b/docs/music.md
@@ -330,11 +330,11 @@ s = synth.OscSynth(wave=amy.PCM, preset=50)
 s.note_on(60, 1.0)
 ```
 
-You can also load PCM patches with looped segments if you have their loopstart and loopend parameters (these are often stored in the WAVE metadata. If the .wav file has this metadata, we'll parse it. The example file `/sys/ex/vlng3.wav` has it. You can also provide the metadata directly.) To indicate looping, use `wave_submode=amy.PCM_LOOP` (or `PCM_LOOP_FOREVER` to keep the loop going while the ADSR fades it out).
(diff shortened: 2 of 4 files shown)


bwhitman:
Pushed b8fcf79 — the one shipped example that used `feedback` for PCM looping:

- `tulip/fs/tulip/ex/bunny_bounce/bunny_bounce.py:211` — `amy.send(osc=0, wave=amy.PCM, preset=10, vel=0.25, feedback=1)` → `mode=amy.PCM_LOOP`. Its frame callback retriggers osc 0 every frame and never sends a note-off, so `PCM_LOOP` preserves the behavior — it's also the migration amy#1006 applies to this exact call shape in its own `test.py`.

I swept the rest and there's nothing else: the remaining `feedback` hits in tulipcc are legitimately non-PCM (echo FX in `amy_parameters.js`/`online.md`, KS in `web/static/examples.js:187`, FM in the DX7 docs), the AMYboard sketches (`sampler.py`, `wavfile.py`) use plain non-looping PCM, and the sketch-generator prompt in `amyboardworld_db_api.py` never taught looping-via-feedback. I also grepped for it hidden in raw patch strings (a `b` on a `w7` osc) — none.

---

Two things in the `music.md` hunk look like they may be slips, flagging rather than changing since they're yours:

1. **`amy.load_sample("/sys/ex/vlng3.wav", patch=50)`** — `load_sample`'s signature is `(wavfilename, preset=0, midinote=0, loopstart=0, loopend=0)`, unchanged by amy#1006, so `patch=` raises `TypeError`. (Confusingly `unload_sample(patch=0)` *does* take `patch` — that asymmetry is probably what's behind it. `docs/tulip_api.md` has the same `patch=`-to-`load_sample` bug in three places already, if you want to sweep those too.)
2. **`synth.OscSynth(wave=amy.CUSTOM, patch=50, ...)`** — `CUSTOM` is 21 (the user-C-DSP wave), `PCM` is 7; and `patch=` is wire `K` (a synth patch) while the loaded sample lives in PCM preset `p`. amy#1006's own docs and tests keep `wave=amy.PCM, preset=N` for this.

So I think that block wants to stay `preset=50` / `wave=amy.PCM` and only change `feedback=1` → `mode=amy.PCM_LOOP`. Happy to push that if you agree — or if the `patch=` rename is something coming in a follow-up, say the word and I'll leave it and fix `tulip_api.md` to match instead.

Note this PR can't merge until amy#1006 lands and the pin moves: `amy.PCM_LOOP` doesn't exist at the currently pinned amy (0454df9), so both the doc and this example reference a constant that isn't there yet.

dpwe:
Please fix the two issues you identified in `music.md`, and the bugs in `tulip_api.md`.  Also, please update `unload_sample` to use `preset=`, not `patch` (and fix any references in docs).

bwhitman:
All three done.

**`music.md`** (b155b6d) — `load_sample(patch=50)` → `preset=50`, and `wave=amy.CUSTOM` → `wave=amy.PCM`. Your `mode=amy.PCM_LOOP` is untouched. Also reworded the sentence to say "PCM presets" rather than "PCM patches" to match.

**`tulip_api.md`** (b155b6d) — the same slip in five places: the cowbell `OscSynth(patch=10)`, both `load_sample(..., patch=50)` calls, the `amy.send(osc=20, wave=amy.PCM, patch=50, ...)` that plays the loaded sample back, and the `unload_sample` example (which also implied a bare `patch` variable was in scope — now `amy.unload_sample(50)`), plus the surrounding prose.

**`unload_sample(preset=)`** — shorepine/amy#1008, since it's an amy change. Kwarg renamed, the unload message it builds is unchanged, and the confirmation print now says `Preset` rather than `Patch`. Nothing in either tree called it by keyword (docs use the positional form), and amy's own `docs/synth.md` already said "preset" throughout, so no amy doc changes were needed. `make test`: 120 pass. Kept it as its own PR rather than folding it into amy#1006 so that one stays about the loop modes — happy to squash them if you'd rather.

I ran the corrected snippets against a desktop build rather than eyeballing them, since that's what would have caught this originally:

```
wav-present | PASS music.md load_sample(preset=) | PASS music.md OscSynth(wave=amy.PCM, preset=)
| PASS tulip_api OscSynth preset=10 (cowbell) | PASS tulip_api load_sample(midinote/loop/preset)
| PASS tulip_api amy.send(wave=amy.PCM, preset=) | PASS tulip_api unload_sample(50) positional
| PASS unload_sample(preset=51) keyword | PASS load_sample(patch=) rejected (TypeError)
| PASS unload_sample(patch=) rejected (TypeError) | PCM=7 CUSTOM=21
```

The last three lines are the confirmation the bugs were real: both `patch=` forms now raise `TypeError`, and `CUSTOM` (21) really is a different wave from `PCM` (7).

Merge order is unchanged — this still needs amy#1006 (for `amy.PCM_LOOP`) and now amy#1008 to land, then a pin bump. Say the word once they're merged and I'll do the pin + `sync_amy_docs.py` + `gen_amy_kwmap.py` refresh and re-verify.

PR #1263, opened 2026-07-29
Add PCM mode parameter for sample looping control

bwhitman:
## Summary
Updates AMY documentation and adds support for the `mode` parameter to control PCM sample playback behavior, replacing the previous `feedback` parameter. This enables fine-grained control over looping behavior in sampled instruments.

## Changes
- **Documentation (`synth.md`)**: Updated PCM looping examples to use the new `mode` parameter instead of `feedback`, with clearer examples demonstrating:
  - `PCM_LOOP`: loops while respecting loop marks, applies ADSR release on note-off
  - `PCM_LOOP_FOREVER`: continuous looping that ignores note-offs (for use with envelope generators)
  - Default behavior: plays through without looping
  
- **Keyword map (`amy_kwmap.h`)**: Added `mode` parameter (`ww` wire format, integer type) to the AMY keyword map, increasing keyword count from 64 to 65

- **API documentation (`api.md`)**: Added `mode` parameter documentation explaining PCM-specific modes and their behaviors

- **Submodule pin**: Updated `amy` submodule to commit `39f724fb80c7` which implements the `mode` parameter support

- **Vendored docs**: Refreshed `_VENDORED_FROM.txt` to reflect the new AMY commit

## Implementation Details
The `mode` parameter provides wave-specific playback modes. For PCM waves, it controls whether samples loop, whether they respect embedded loop marks, and how they respond to note-off events. This replaces the simpler boolean `feedback` approach with more granular control suitable for different instrument types (e.g., looping strings vs. one-shot percussion).

https://claude.ai/code/session_019fjnuccb1u4ZitLb8Sk5Zo


bwhitman:
Yes — it's in this PR (`873d42ff`), regenerated at the new pin:

```
- * Source of truth: amy/amy/__init__.py _KW_MAP_LIST at 0454df9d311d
- * (64 keywords).
+ * Source of truth: amy/amy/__init__.py _KW_MAP_LIST at 39f724fb80c7
+ * (65 keywords).
+    {"mode", "ww", 'I'},
-#define AMY_KW_N 64
+#define AMY_KW_N 65
```

Worth being precise about what was and wasn't broken, since it's easy to assume the worse case:

**On main today it was never stale.** main's pin is still the old amy (`0454df9`), where `mode` doesn't exist, so its 64-keyword table matches its own pin exactly — I verified `gen_amy_kwmap.py --check` passes against `origin/main` as it stands. The header only had to change *because* the pin moves, which is what this PR does, and `amy-pin-check` is what would have caught it if I'd forgotten (it's green here, along with `refdocs-fresh`).

**And the failure mode if it had been missed is mild by construction.** A keyword absent from the table makes `tulip.amy_message()` return `None` and fall back to the Python builder — right answer, just no longer the fast path. So a stale kwmap costs speed on `mode=` messages, never correctness. That's the whole point of bailing rather than guessing.

Verified at this pin on a desktop build: all five sub-modes build byte-identically to the Python builder through C — `PCM_LOOP` → `v0w7l0.25p10ww2Z` — and the corpus is at 0 mismatches with the `mode=` cases added.

One thing to flag while this sits: main right now has `bunny_bounce.py:211` and the docs calling `amy.PCM_LOOP` against the pre-`mode` pin, so that line raises `AttributeError` on a build from main until this lands. Nothing's shipped from that state, but it's the reason to merge this sooner rather than later.

PR #1297, opened 2026-08-10
Make callbacks GC roots

adamschabtach:
# Make callbacks GC roots

## The bug

Every callback Tulip stores is a plain C global, and none is declared as a GC
root:

| Global | File |
|---|---|
| `midi_callback`, `amy_overload_callback` | `tulip/shared/modtulip.c:106,116` |
| `frame_callback`, `frame_arg`, `touch_callback`, `keyboard_callback`, `ui_quit_callback`, `ui_switch_callback` | `tulip/shared/modtulip.c:1139-1144` |
| `sequencer_callbacks[]`, `defer_callbacks[]`, `defer_args[]` | `tulip/shared/tsequencer.c:5,9,10` |

`MICROPY_PORT_ROOT_POINTERS` does not appear anywhere in the tree, and
`gc_collect()` on the esp32 port scans **registers and the stack only**
(`ports/esp32/gccollect.c`). A C global is therefore invisible to the collector.

So this — the ordinary way to use the API, and what `tulip/fs/tulip/ex/`,
`shared/py/voices.py` and the docs all do — is a use-after-free:

```python
tulip.touch_callback(self.on_touch)
```

`self.on_touch` builds a **new bound-method object** on every evaluation.
Handed straight over, the C global holds the only reference to an object
nothing else can reach, so a collection frees it. The pointer survives; the
object does not.

Whether anything then breaks depends on **when that block is reallocated**,
which is what makes this so unpleasant to diagnose:

- silent — the pointer is still non-NULL, nothing raises;
- delayed — it needs a collection *and* a reuse;
- unreproducible — it depends on heap churn;
- and re-registering the same handler fixes it instantly, because that
  makes a fresh object.

I hit this as a touchscreen that died after **3.5 hours** of a running
sequencer, with no error logged, on a Tulip CC v4r11. Re-registering fixed it
immediately every time. I had it mitigated with a 30-second re-assertion for
weeks before reading the source. It is also the most plausible explanation for
a one-off spontaneous reset I never accounted for: scheduling a freed object.

Note that the sequencer callbacks have the same defect, so a long-running
`seq_add_callback` is on the same footing — it simply tends to get luckier
about reallocation.

## Reproduction

⚠️ This may hang or reset the device. That is the bug.

```python
import gc, time, tulip

fired = []

def install():
    # A closure whose ONLY reference will be tulip's C global.
    def cb(*a):
        fired.append(1)
    tulip.frame_callback(cb)

install()
time.sleep(1)
print("firing before gc:", len(fired) > 0)

n = len(fired)
gc.collect()
# Churn the heap so the freed block gets reused.
junk = [bytearray(96) for _ in range(3000)]
time.sleep(1)
print("still firing after gc + churn:", len(fired) > n)
```

Without the patch the callback object is collectable at `gc.collect()` and the
allocation loop reuses its memory. With the patch it stays reachable.

## The fix

Register each with `MP_REGISTER_ROOT_POINTER` and access through
`MP_STATE_PORT`, matching the existing use in `proxy_c.c` and for
`native_code_pointers`. Same globals, same NULL semantics, same call sites —
only their storage moves into the root-pointer section.

The array sizes have to be literals: `MP_REGISTER_ROOT_POINTER` is collected
into `genhdr/root_pointers.h`, which `mpstate.h` includes long before
`tsequencer.h` is visible. `MP_STATIC_ASSERT` keeps them honest against
`SEQUENCER_SLOTS` and `DEFER_SLOTS`.

## What I have and have not verified

**Verified:** the analysis above, from source at the commit my board runs
(`a2c981c4`) and the pinned MicroPython (`82e69df33`). And the mechanism, on
hardware: holding a strong Python-side reference to every callback made the
failure go away, confirmed across 40 forced `gc.collect()` calls with ticks,
frames, touch and MIDI all still live afterwards.

**Not verified:** I have not compiled this. I do not have an ESP-IDF toolchain
set up, so the patch is offered for CI and review rather than as something I
have run. It is mechanical and I have checked every call site converted, but
please treat the build as unproven.

## Related, deliberately not fixed here

These globals are also **not cleared on soft reset** — `tsequencer_init()` is
called once from `app_main`, outside the `soft_reset:` loop, and nothing zeroes
the root-pointer section. So a callback registered before a soft reset survives
as a dangling pointer into a heap that has since been re-created. That is
pre-existing and orthogonal; this PR does not change it either way, but it
probably wants its own fix.

## Workaround for anyone hitting this today

Keep your own reference to the exact object you hand over:

```python
self._cb_touch = self.on_touch          # reachable from something live
tulip.touch_callback(self._cb_touch)
```

Not `tulip.touch_callback(self.on_touch)`, which creates an object only the C
global will hold.

### DIFF
--- a/tulip/shared/amy_connector.c
+++ b/tulip/shared/amy_connector.c
@@ -49,8 +49,6 @@ void midi_out(uint8_t * bytes, uint16_t len) {
 // A queue to store the AMY midi messages coming IN
 uint8_t last_midi[MIDI_QUEUE_DEPTH][MAX_MIDI_BYTES_PER_MESSAGE];
 uint8_t last_midi_len[MIDI_QUEUE_DEPTH];
-extern mp_obj_t midi_callback;
-extern mp_obj_t amy_overload_callback;
 
 int16_t midi_queue_head = 0;
 int16_t midi_queue_tail = 0;
@@ -59,8 +57,8 @@ int16_t midi_queue_tail = 0;
 // (it has already reset the synth and played its bleep). Hand off to Python
 // with the load as a percent -- a small int, so no cross-task heap allocation.
 void tulip_amy_overload_hook(float load) {
-    if (amy_overload_callback != NULL)
-        mp_sched_schedule(amy_overload_callback, MP_OBJ_NEW_SMALL_INT((mp_int_t)(load * 100.0f)));
+    if (MP_STATE_PORT(amy_overload_callback) != NULL)
+        mp_sched_schedule(MP_STATE_PORT(amy_overload_callback), MP_OBJ_NEW_SMALL_INT((mp_int_t)(load * 100.0f)));
 }
 
 
@@ -226,7 +224,7 @@ void tulip_midi_input_hook(uint8_t * data, uint16_t len, uint8_t is_sysex) {
             }
             sysex_len = len;
         }
-        if(midi_callback!=NULL) mp_sched_schedule(midi_callback, mp_const_true);
+        if(MP_STATE_PORT(midi_callback)!=NULL) mp_sched_schedule(MP_STATE_PORT(midi_callback), mp_const_true);
     } else {
         for(uint32_t i = 0; i < (uint32_t)len; i++) {
             if(i < MAX_MIDI_BYTES_PER_MESSAGE) {
@@ -243,7 +241,7 @@ void tulip_midi_input_hook(uint8_t * data, uint16_t len, uint8_t is_sysex) {
         }
 
         // We tell Python that a MIDI message has been received
-        if(midi_callback!=NULL) mp_sched_schedule(midi_callback, mp_const_false);
+        if(MP_STATE_PORT(midi_callback)!=NULL) mp_sched_schedule(MP_STATE_PORT(midi_callback), mp_const_false);
     }
 }
 
--- a/tulip/shared/keyscan.c
+++ b/tulip/shared/keyscan.c
@@ -426,20 +426,19 @@ uint16_t scan_ascii(uint8_t code, uint32_t modifier) {
 }
 
 extern int16_t lvgl_is_repl;
-extern mp_obj_t keyboard_callback, ui_quit_callback, ui_switch_callback;
 
 void send_key_to_micropython(uint16_t c) {
     // handle the global system hotkeys before anything else. we have two, ctrl-tab and ctrl-q 
     if(c==17) {
-        if(ui_quit_callback != NULL) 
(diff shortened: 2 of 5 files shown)


bwhitman:
Can you bump this to current main and also fix the tsequencer.c issue?

adamschabtach:
Rebased onto current main.

On the tsequencer.c issue — the build has been failing since I opened this, and the cause was mine:

```
tulip/shared/tsequencer.c:44:1: error: expected ')'
   44 | MP_STATIC_ASSERT(SEQUENCER_SLOTS == 8);
      | ^
../../micropython/py/misc.h:54:39: note: expanded from macro 'MP_STATIC_ASSERT'
   54 | #define MP_STATIC_ASSERT(cond) ((void)sizeof(char[1 - 2 * !(cond)]))
```

`MP_STATIC_ASSERT` expands to an *expression*, so it is only valid inside a function body — I had the two asserts at file scope, and `tsequencer.o` never compiled.

Those asserts were guarding a premise that turns out to be false anyway. I had written the root-pointer array sizes as literals (`[8]`, `[32]`) because `genhdr/root_pointers.h` is pulled in by `mpstate.h` long before `tsequencer.h` is visible. But `MP_REGISTER_ROOT_POINTER` is collected from **preprocessed** source (`SRC_QSTR` → `qstr.i.last` → `make_root_pointers.py`, see `py/mkrules.mk`), so `#define`s expand on the way in. It is the same reason the generated header already reads `readline_hist[(50)]` and `sched_queue[(128)]`.

So the declarations now just use the names:

```c
MP_REGISTER_ROOT_POINTER(mp_obj_t sequencer_callbacks[SEQUENCER_SLOTS]);
MP_REGISTER_ROOT_POINTER(mp_obj_t defer_callbacks[DEFER_SLOTS]);
MP_REGISTER_ROOT_POINTER(mp_obj_t defer_args[DEFER_SLOTS]);
```

and both static asserts and the comment above them are gone. The generated header still comes out as

```c
mp_obj_t sequencer_callbacks[8];
mp_obj_t defer_callbacks[32];
mp_obj_t defer_args[32];
```

so the roots now track `tsequencer.h` by construction — which is what the asserts were trying to buy in the first place.

Verified against a local desktop build (`tulip/macos`, `make DEBUG=1`): compiles and links clean, and the generated `root_pointers.h` is as above. Nothing else moved in the rebase — I diffed the patch against its pre-rebase self, and this `tsequencer.c` change is the only difference.

If "the tsequencer.c issue" meant something other than the build break, say the word and I'll fix that too.

bwhitman:
thank you!!!!

PR #1300, opened 2026-08-10
Fix dropdown scroll flicker on hardware by making the open list cover

bwhitman:
## Problem

On Tulip hardware, scrolling the open Patch dropdown in the juno app flickers violently — the sliders underneath flash through the list while your finger moves. Web and desktop are unaffected.

## Cause

Two things intersect:

1. The default LVGL theme gives a dropdown's open list rounded corners **and** `clip_corner = true`, and `clip_corner` makes LVGL's cover check return `MASKED` unconditionally (`lv_obj.c`). The list therefore never counts as covering the screen area under it, so every scroll tick redraws the whole invalidated rect bottom-up: background, then the juno sliders, then the list painted over them.
2. On hardware, LVGL renders in `LV_DISPLAY_RENDER_MODE_DIRECT` straight into `bg` — the same single buffer the ESP32-S3 RGB panel is continuously DMA-scanning via bounce buffers — with a no-op flush callback and no vsync synchronization. The scan-out repeatedly catches the "sliders drawn, list not yet" intermediate state, which reads as the whole dropdown flickering. Desktop/web snapshot the finished `bg` and present whole frames, so the intermediate states are never visible there.

## Fix

Add `tulip.lv_dropdown_opaque()` (in `ui.py`): square corners, no corner clipping, opaque bg on the dropdown's list, so it passes the cover check and scroll redraws start *at the list*, never touching the sliders. Applied to both `JunoDropDown` (patch + channel selectors in juno6) and the drums tone dropdown.

Trade-off: the open list loses its rounded corners. Remaining artifact is mild text shear while scrolling (inherent to unsynced direct-mode rendering), which is far less objectionable than the flash-through.

## Testing

- Verified the exact LVGL API surface at runtime against Tulip Desktop's REPL: `get_list()` + the three style setters apply cleanly (`radius=0`, `clip_corner=False`, `bg_opa=255`).
- Frozen-module change, so it needs a firmware build to test on device; flicker repro is hardware-only.

### DIFF
--- a/tulip/shared/display.c
+++ b/tulip/shared/display.c
@@ -36,10 +36,12 @@ uint16_t *sprite_h_px;//[SPRITES];
 uint8_t *sprite_vis;//[SPRITES];
 uint32_t *sprite_mem;//[SPRITES];
 
-// LVGL renders into this shadow of bg; lv_flush_cb_8b blits finished areas into bg.
-// LVGL can't render directly into bg: on hardware the RGB panel scans bg out
-// continuously, so mid-render widget states would be visible as flicker.
+// LVGL renders into this small band buffer (PARTIAL mode); lv_flush_cb_8b blits
+// each finished band into bg. LVGL can't render directly into bg: on hardware the
+// RGB panel scans bg out continuously, so mid-render widget states would be
+// visible as flicker.
 uint8_t * lv_buf;
+#define LV_BUF_BYTES (((H_RES+OFFSCREEN_X_PX)*(V_RES+OFFSCREEN_Y_PX)*BYTES_PER_PIXEL)/10)
 
 uint8_t *TFB;//[TFB_ROWS][TFB_COLS];
 uint8_t *TFBfg;//[TFB_ROWS][TFB_COLS];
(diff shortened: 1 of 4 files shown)


bwhitman:
Round 2: the styling change wasn't enough — the real culprit is that LVGL renders in DIRECT mode straight into `bg`, the buffer the RGB panel is continuously scanning out. A large redraw (the open patch list during a drag) takes several ms of software rendering, so the scan repeatedly displays half-painted widgets regardless of what's underneath.

New commit renders LVGL into a shadow buffer (`lv_buf`, same geometry as `bg`, +~800KB SPIRAM on TULIP4) and blits each *finished* area into `bg` in the flush callback. The scan-out never sees a partial render anymore; worst case is a brief shear between two complete frames. This fixes LVGL flicker on hardware globally (keyboards, launcher, etc.), not just this dropdown.

Verified on a local macOS desktop build via the stdin REPL harness: an opaque red LVGL rect reads back `224` (RGB332 pure red) from `bg` at its exact coordinates through the new flush path, and an opened dropdown's list content lands in `bg` too. Needs hardware to confirm the flicker is gone.

bwhitman:
Round 3 (`8b76f344`): same flicker fix, ~10x less memory. The render buffer is now a 1/10-screen band (~80KB SPIRAM) in `LV_DISPLAY_RENDER_MODE_PARTIAL` instead of the full-frame shadow (~800KB). LVGL renders each invalidated area in bands and the flush cb blits each finished band into `bg` — the scan-out still never sees a partial render.

Archaeology: this is actually a **return to the original architecture**. Until 676ba9c6 (Apr 2024, "starting to get rgb332 working") LVGL rendered RGB565 into a small `lv_buf` and the flush cb converted+copied into `bg`. That commit switched to DIRECT mode straight into `bg` (the copy became free once the render format matched bg's RGB332) — which silently introduced the scan-out race. That's when hardware got flickery.

Verified DIRECT↔PARTIAL behavior parity on desktop (bg_pixel readback): full-screen invalidate repaints tulip-bg-drawn pixels with the opaque screen color in **both** modes, local invalidates leave distant bg pixels untouched in both, widget pixels land identically. Waiting on hardware confirmation + free-heap readback (should now only drop ~80KB vs stock).

PR #1347, opened 2026-09-06
amyboard CV: sample both ADS1015 channels as a tightly-spaced pair

dpwe:
## Problem

Reports of spurious reads on the CV input (ADS1015). Two separate issues, both in `cv_read_task`'s ADC path.

**`tulip.cv_in` and AMY's coef hook are *not* conflicting** — worth stating up front, since that was the initial suspicion. Since the `cv_read_task` refactor both paths are pure reads of a cached value (`cv_input_hook`, no I2C), so the audio thread and the MicroPython task cannot step on each other at the ADC. Only `cv_read_task` talks to the device.

### 1. ~5 ms of skew between ch0 and ch1

`read_ads1015_raw()` speculatively started the *other* channel's conversion at the end of each call. So ch0's conversion ran at the end of the previous iteration and then sat in the CONVERT register across the whole 6 ms `xTaskDelayUntil`, while ch1's was started microseconds before its own read:

| | old | new |
|---|---|---|
| ch0 sample age at use | ~6 ms | ~1.5 ms |
| ch1 sample age at use | ~0.3–1.3 ms | ~0.3 ms |
| **inter-channel skew** | **~5 ms** | **~1.25 ms** |

Structural, not jitter — close to the worst arrangement available. Anything comparing CV0 against CV1 (a stereo pair, an X/Y gesture) saw the two channels disagree.

The ADS1015 has one mux and one delta-sigma modulator, so simultaneous sampling is not available at any price, and each reading integrates over its own ~303 µs conversion window. What this buys is two *adjacent* windows instead of two windows 5 ms apart.

### 2. Every I2C result along the path was discarded

- **`ads1015_start_conversion()` set `ads1015_pending_channel` unconditionally.** A NAKed or timed-out config write left the code believing the mux had moved when it had not; the next read then skipped starting a conversion and returned *the other channel's* sample, until some later write happened to succeed. **This is the most likely cause of the reported symptom** — it is intermittent and correlates with bus contention.
- **`ads1015_read_register()` discarded the register-pointer write's result.** A failed pointer-set with a successful read returned whatever register the pointer was last left at: CONFIG's ~0x85E3 read as a sample, or a negative sample's sign bit read as OS-ready.
- **`ads1015_get_result()` discarded the CONVERT read's result** and returned the buffer regardless; `read_ads1015_raw()` returned a bare `uint16_t` with no way to signal failure, so `cv_read_task` always cached *something*.

## Changes

**Tight-pair sequencing.** Both conversions run back to back in one iteration, waited out by yielding (`vTaskDelay(1)` plus OS-bit confirmation, spinning out only the rare early wake) so the task burns no CPU. Spinning ch0's conversion instead would give ~0.7 ms skew at the cost of 320 µs busy-wait per 6 ms period — 5% of core 0 — which is not worth it against the ~5 ms we started from.

**The cached channel is gone.** Every conversion is started explicitly with its result checked, so there is no cached notion of which channel the mux is on that can drift out of step with the hardware. The desync bug is structurally impossible rather than patched.

**Errors propagate; the last good pair is held** instead of publishing a value we cannot vouch for.

**Repeated-START register reads.** The pointer write and the read are now one transaction: 12 acquisitions of the legacy driver's per-port mutex per pair read drop to 7, the bus is never released mid-read, and the two halves can no longer disagree. The last byte of a read is now NACKed rather than ACKed — ACKing it tells the device to load another byte, so it can still be driving SDA when the STOP goes out, which is a classic source of intermittent corruption on the *next* transaction.

**Coherent publish.** `cv_cached_value[]` becomes a double-buffered pair published with a single index store, so a reader sees a snapshot of both channels rather than ch0 from one update against ch1 from the next.

**`CV_READ_TASK_PRIORITY` `PRIO_MIN+1` to `+2`**, so an OLED flush (`i2c_bg_task`, same priority, same core) cannot preempt the task mid-pair. The I2C follower at priority 20 still outranks it.

**`tulip.cv_stats()` returns `(errors, last_skew_us, max_skew_us)`**, any argument resets. This is how to tell a genuinely failing ADC (errors climbing) from scheduling jitter (skew outliers), and how to check the real skew rather than my arithmetic.

Overriding both channels via `tulip.cv_local()` still takes the ADC out of the loop entirely, which hwci's `no_cv` A/B run depends on.

## Testing

Builds clean for `AMYBOARD` (ESP-IDF 5.4.1), no warnings. `tulip.cv_stats()` is inside the existing `#ifdef AMYBOARD` / `#ifdef ESP_PLATFORM` nesting, so the TULIP4_R11 build is untouched.

**Not yet verified on hardware.** The timings here are arithmetic, and the repeated-START and NACK changes alter the wire protocol — this wants a board before merging. On one, `tulip.cv_stats()` should show skew near the ~1.25 ms estimate and `errors` staying at 0; if `errors` climbs instead, the original spurious reads were bus/device failures rather than the `pending_channel` desync, and that is a different fix.

The NACK fix is separable from the repeated-START change if it is worth bisecting them on hardware.

### DIFF
--- a/tulip/amyboard/amyboard_support.c
+++ b/tulip/amyboard/amyboard_support.c
@@ -19,6 +19,8 @@
 #include "freertos/event_groups.h"
 #include "esp_log.h"
 #include "esp_err.h"
+#include "esp_rom_sys.h"   // esp_rom_delay_us for the sub-tick ADC spin remainder
+#include "esp_timer.h"     // esp_timer_get_time for the CV pair skew measurement
 #include "driver/i2s_std.h"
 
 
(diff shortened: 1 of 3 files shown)


dpwe:
---

## Verified on hardware (AMYboard, ESP32-S3 rev v0.2)

Flashed to a board and measured with `tulip.cv_stats()`, 400 samples each, idle and under load (8-voice piano patch 256 + Python `i2c.scan()` every 25 samples).

### The first version of this PR was wrong, and the board said so

Yielding both conversion waits cost exactly the skew this change was meant to buy. `vTaskDelay()` returns when the task is next *scheduled*, not at the tick edge — at `PRIO_MIN+2` on core 0 that means waiting out AMY's render task (`PRIO_MAX-1`, same core, every 5.8 ms), so the skew clustered at block-period multiples.

| skew | idle median | loaded median | loaded p90 | loaded max |
|---|---|---|---|---|
| yield both waits | 1497 µs | **4922 µs** | 5506 µs | 9852 µs |
| **spin ch0's wait** (`f30fba0`) | **722 µs** | **736 µs** | 4780 µs | 8582 µs |
| *(before this PR: ~5 ms, structural)* | | | | |

Yielding both was no better under load than the speculative-start scheme it replaced. ch0's wait is inside the skew window so it now spins — 320 µs per 6 ms period, ~5% of core 0, which is otherwise idle at this task's priority. ch1's wait is outside the window and still yields.

### Error paths never fire in normal operation

`errors == 0` across 10 s idle, under the piano load, and with Python hammering the bus with `i2c.scan()`. `cv_in()` reads sane volts on unconnected jacks (−0.03 V, +0.33 V).

### Two things found that are NOT caused by this PR

1. **~445 I2C errors during boot.** `cv_read_task` is created in `main.c`, but `I2C_NUM_0`'s master driver isn't installed until Python calls `machine.I2C(0)` in `amyboard.py:783` — so the first ~2.7 s of reads fail with `i2c driver not installed` on the console. Pre-existing; the old code did it too (louder — ~10 errors per iteration rather than 1). This PR is what makes it *visible*, by counting it. Worth fixing separately.
2. **Residual p90 of ~4.8 ms under load** is render-task preemption landing between the two conversions. Bounding it would mean rejecting pairs whose measured skew exceeds a threshold — separate change.

dpwe:
I tested this on a real AMYboard with an OLED with a few display-heavy scripts (e.g, live_oscilloscope-2) and everything seems good.

Issue #32, opened 2023-03-16
timer crash


bwhitman:
```python
import machine
c = 0
def handle(timer):
    global c
    c+=1
timer = machine.Timer(0)
timer.init(period=1000, mode=machine.Timer.PERIODIC,callback=handle)
```

crashes with

```
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x40384448  PS      : 0x00060033  A0      : 0x82008826  A1      : 0x3fc9b8b0  
0x40384448: vTaskGenericNotifyGiveFromISR at /Users/bwhitman/outside/tulipcc/esp-idf/components/freertos/tasks.c:5818

A2      : 0x00000000  A3      : 0x00000000  A4      : 0x3fc9b8d0  A5      : 0x3fc9b950  
A6      : 0x00060423  A7      : 0x00000003  A8      : 0x00000168  A9      : 0x00000168  
A10     : 0x00000002  A11     : 0x3fc9e6e4  A12     : 0x00060023  A13     : 0x00060023  
A14     : 0x00000004  A15     : 0x0000abab  SAR     : 0x00000020  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000168  LBEG    : 0x00000000  LEND    : 0x00000000  LCOUNT  : 0x00000000  


Backtrace:0x40384445:0x3fc9b8b00x42008823:0x3fc9b8d0 0x420088cf:0x3fc9b900 0x40376c46:0x3fc9b920 0x420f85eb:0x3fcf5af0 0x4200322f:0x3fcf5b10 0x40382898:0x3fcf5b30 
0x40384445: vTaskGenericNotifyGiveFromISR at /Users/bwhitman/outside/tulipcc/esp-idf/components/freertos/tasks.c:5818

0x42008823: mp_hal_wake_main_task_from_isr at /Users/bwhitman/outside/tulipcc/ports/esp32s3/mphalport.c:198

0x420088cf: machine_timer_isr at /Users/bwhitman/outside/tulipcc/ports/esp32s3/machine_timer.c:169

0x40376c46: _xt_lowint1 at /Users/bwhitman/outside/tulipcc/esp-idf/components/freertos/port/xtensa/xtensa_vectors.S:1111

0x420f85eb: cpu_ll_waiti at /Users/bwhitman/outside/tulipcc/esp-idf/components/hal/esp32s3/include/hal/cpu_ll.h:182
[...]

PR #1361, opened 2026-09-20
amyboard: assign mp_main_task_handle so ISRs don't notify a NULL task

mateusz28011:
Any `machine.Pin.irq` that actually fires resets an AMYboard. Same root cause as #32. Repro on `main`
(af772c83), on an **unconnected** pin — the internal pull makes the edge:

    import machine
    p = machine.Pin(16, machine.Pin.IN, machine.Pin.PULL_DOWN)
    p.irq(trigger=machine.Pin.IRQ_RISING, handler=lambda pin: None)   # board still fine here
    p.init(machine.Pin.IN, machine.Pin.PULL_UP)                       # one rising edge -> USB drops
    # after the reboot: machine.reset_cause() == machine.HARD_RESET

Calling the handler from the REPL or through `micropython.schedule` is fine, so it is the ISR path.

**Cause.** `machine_pin_isr_handler` ends in `mp_hal_wake_main_task_from_isr()`, i.e.
`vTaskNotifyGiveFromISR(mp_main_task_handle, ...)` (`tulip/amyboard/mphalport.c:264`). Stock MicroPython
passes `&mp_main_task_handle` to `xTaskCreatePinnedToCore`; `tulip/amyboard/main.c:355` passes
`&tulip_mp_handle`, and `mp_main_task_handle` has no writer anywhere under `tulip/`. The ISR notifies NULL.

I have no backtrace (console is USB CDC, the panic goes to UART0), but #32's shows this exact frame with
`A2 = 0`: `vTaskGenericNotifyGiveFromISR` ← `mp_hal_wake_main_task_from_isr`, LoadProhibited at 0x168.
#32 was closed by removing the wake calls; since `b1d016c4` the build uses stock `machine_pin.c` /
`machine_timer.c`, so the calls are back and the handle is still unassigned. `machine_uart.c` uses the
same function, and in-tree `tulip/shared/py/sx126x.py` uses `Pin.irq`.

**Fix.** One line in `tulip/amyboard/main.c` after `xTaskCreatePinnedToCore`:
`mp_main_task_handle = tulip_mp_handle;`. (Passing `&mp_main_task_handle` to the create call and aliasing
`tulip_mp_handle` would also close the theoretical window before the assignment — say if you prefer it.)

**Tested on hardware**, an AMYboard running af772c83 + this commit: the repro above no longer resets and
the handler runs. Flipping the pull 200 times on unconnected IO16 gave 207 callbacks (floating pin, a few
extra edges expected) with no reset. The same with a rising-edge IRQ on IO14 and a falling-edge IRQ on
IO15 armed together, 50 pull flips each: 55 / 50 callbacks, no reset, audio kept running.

Not included: the same line in `tulip/esp32s3/main.c:440`. #32's backtrace is from `ports/esp32s3`, so it
is the same bug there, but I have no Tulip CC to test on — glad to add it.


bwhitman:
thank you !

